PR #183, opened 2018-10-16
Small fixes 1

revolunet:
minor fixes/cleanup


packages/code-du-travail-frontend/package.json
```diff
@@ -8,7 +8,8 @@
   "main": "src/index.js",
   "dependencies": {
     "@sentry/browser": "^4.1.0",
-    "@socialgouv/code-du-travail-ui": "^1.2.3",
+    "@socialgouv/code-du-travail-css": "^1.0.2",
+    "@socialgouv/code-du-travail-ui": "^1.1.0",
```
lionelB: why not the `^1.2.3` version ?

packages/code-du-travail-frontend/src/Categories.js
```diff
@@ -3,7 +3,7 @@ import { Link } from "../routes";
 
 import { Container, Categories } from "@socialgouv/code-du-travail-ui";
 
-import themes from "./data/themes2";
+import themes from "./data/themes";
```
lionelB, line 6: this will probably confilct with my renaming to `json` 
will see

lionelB (approved):
Remove! Remove! remove!

packages/code-du-travail-frontend/package.json
revolunet: totally :)

revolunet: i will move that into https://github.com/SocialGouv/code-du-travail-numerique/pull/176

PR #219, opened 2018-10-23
WIP: Dockerify2

revolunet:
Quelques fixes :
 - un probleme d'url dans SearchResults (missing `/`prefix) ?
 - une typo dans l'instantiation piwik
 - correction des routes pour les conventions collectives (en attente de unify-routes)

Et surtout un setup docker+lerna qui fonctionne mais est d'une lenteur extrême.

 - le `.env` et les `Dockerfile` sont à la racine


Dockerfile.python
```diff
@@ -16,12 +19,8 @@ RUN apk update \
...
-
-WORKDIR $APP_DIR
+COPY ./packages/code-du-travail-data/Pipfile .
 
 RUN pipenv install --deploy
 
-COPY . $APP_DIR
-
-ENTRYPOINT ["/app/docker/entrypoint.sh"]
+COPY ./packages/code-du-travail-data .
```
lionelB, line 26: le fichier docker/python-entrypoint.sh n'est plus appelé ?

packages/code-du-travail-api/src/server/index.js
```diff
@@ -1,38 +1,39 @@
 // Set up Elastic APM agent: the agent must be started before any other modules.
 // https://www.elastic.co/guide/en/apm/agent/nodejs/current/koa.html
-require('elastic-apm-node').start({
```
douglasduteil, line 3: Those changes shouldn't be part of this PR

douglasduteil (changes requested):

README.md
```diff
@@ -65,7 +64,18 @@ $ yarn workspace <package-name> test
 
 # For example for "code-du-travail-frontend"
 $ yarn workspace code-du-travail-frontend test
+```
+
+### Prod
+
+Edit `.env`
```
lionelB, line 71: I would say 

```sh 
$ cp .env.sample .env
```
And edit the .env file.

douglasduteil: Isn't is automatique on "postinstall" ?

Dockerfile.frontend
```diff
@@ -0,0 +1,37 @@
...
+# TODO: we should switch to https://github.com/zeit/next.js/tree/master/examples/with-universal-configuration-runtime ?
+ARG API_URL
+ARG INTERNAL_API_URL
+ARG SENTRY_PUBLIC_DSN
+ARG PIWIK_URL
+ARG PIWIK_SITE_ID
+
+ENV NODE_ENV=production
+ENV API_URL=$API_URL
+ENV INTERNAL_API_URL=$INTERNAL_API_URL
```
lionelB, line 30: what 's the difference between `API_URL` and `INTERNAL_API_URL` ?

douglasduteil:
Wait @revolunet why did you close your PR ?

revolunet:
will be back soon !

Dockerfile.frontend
revolunet: yep INTERNAL will not exist but let me explain where it came from as i wasted waaay too much time on this :)

We're currently using next.js which does SSR. 

When the user request some content client-side, like autosuggest, he uses his browser `fetch()` and the public `process.env.API_URL` to resolve the data from our API. all fine.

But when the user request a *page* of content like `/question/how-are-you-today` then that's the server who make the initial request to the API and inject content into the page before returning it to user. this happens in the pages `getInitialProps()` function, which also use `process.env.API_URL`.

So at some point, the server use the public API_URL to make requests to the API server which in fact stands on the same docker-compose environment. 

https://github.com/zeit/next.js/#fetching-data-and-component-lifecycle

revolunet: Also these variables are used at build time and i think we should make them overridable at runtime

Dockerfile.python
revolunet: si je ne m'abuse il ne servait a rien a part poller sur l'elastic search

lionelB:
I've made a PR with getConfig... hope it will return soon :)

PR #286, opened 2018-11-16
test(frontend): include all source files in the coverage result

douglasduteil:


douglasduteil:
The coverage might go low low low low low because of that...
But at least it will be the real coverage of the frontend sources

douglasduteil:
Aïe @lionelB

lionelB (approved):
👍

PR #288, opened 2018-11-16
fix(data): add date in fiche ministere

lionelB:

PR #289, opened 2018-11-16
fix(data): remove convention-collective 2121 from ES

lionelB:


douglasduteil (approved):
Can you remove the first commit that is #288 PR ?

PR #295, opened 2018-11-19
feat(data): add sub dataset module to the monorepo

douglasduteil:
---

Note "packages/code-du-travail-data/dataset/fiches_service_public/package.json" is ignored due to yarn install failure...

---

**BREAKING CHANGE:** add sub dataset module to the monorepo

This might impact the build script as the newest in range dependencies are used now.


packages/code-du-travail-data/dataset/code_du_travail/eposeidon_script/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "build_eposeidon",
+  "name": "@cdt/data...build_eposeidon",
```
lionelB, line 2: why there is `...` in packages name ?

douglasduteil: why not :) 

- `/` si prohibited so we need to find something else as separator
- `-` and `_` are too common

douglasduteil: I made an article about `...` back then :)
https://medium.com/@douglasduteil/npm-scoping-for-free-f162c1db4638

lionelB: > - and _ are too common
😭 I like - and _ 

On a more general note I thought you would move those packages to the monorepo root.

douglasduteil: Maybe under an `@cdt/dataset` namespace 
however moving the files will alter the Python indexation script.
@revolunet do we have kind of Python tests that can help us knowing if the indexation script still work after moving the dataset files ?

revolunet: the create_indexes script itself should fail when it cannot find source files

lionelB (approved):

PR #302, opened 2018-11-21
add test for common

lionelB:
add test for commons


packages/code-du-travail-frontend/src/common/Disclaimer.js
```diff
@@ -5,6 +5,7 @@ const Disclaimer = () => (
     Important ! Ce site est en cours de construction : les données qui s'y
```
douglasduteil: can you use `&apos;` here ?

douglasduteil (changes requested):

lionelB:
poke @douglasduteil

packages/code-du-travail-frontend/src/common/FeedbackForm.js
```diff
@@ -0,0 +1,200 @@
...
+      alert("Merci de compléter le formulaire");
+      return;
+    }
+
+    this.setState(
+      {
+        status: "sending"
+      },
+      () => {
+        fetch(formspreeUrl, {
```
douglasduteil: We shouldn't make HTTP calls inside a component.
You might want to separate this logic from the component rendering.

packages/code-du-travail-frontend/src/common/FeedbackForm.js
```diff
@@ -0,0 +1,200 @@
...
+            disabled={this.state.status === "sending"}
+          >
+            Envoyer ma question
+          </button>
+          <div className="Feedback__status">
+            {this.state.status === "sent" ? (
+              <Alert success>Message bien envoyé !</Alert>
+            ) : this.state.status === "error" ? (
+              <Alert warning>Impossible d&apos;envoyer votre message</Alert>
+            ) : null}
```
douglasduteil: It looks like a "switch/case" to me no ?

packages/code-du-travail-frontend/src/common/FeedbackForm.js
```diff
@@ -0,0 +1,200 @@
...
+  componentWillUnmount() {
+    clearTimeout(this.timeoutId);
+  }
+
+  render() {
+    const { results, query } = this.props;
+    return (
+      <form
+        action={formspreeUrl}
+        className="Feedback__form"
```
douglasduteil: Why "uppercases" in css class names ?

packages/code-du-travail-frontend/src/common/ServiceRenseignement.js
```diff
@@ -0,0 +1,66 @@
...
+            target="_blank"
+            rel="noopener noreferrer"
+            href={this.state.departmentData.url}
+            className="external-link__after"
+          >
+            <mark>{this.state.departmentData.url}</mark>
+          </a>
+        </p>
+      );
+    }
```
douglasduteil: Might want to take a functional approach

```jsx
const department = this.state.departmentData ? DepartmentComponent(this.state.departmentData) : null

[...]

const DepartmentComponent = data => (
  <p>
    <a
      target="_blank"
      rel="noopener noreferrer"
      href={data.url}
      className="external-link__after"
    >
      <mark>{data.url}</mark>
    </a>
  </p>
);
```

Here the `DepartmentComponent` is a standalone component

packages/code-du-travail-frontend/src/common/__tests__/Html.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe("<Html />", () => {
+  it("should render", () => {
+    const { container } = render(<Html>{"<b>Hello</b> World"}</Html>);
+    expect(container).toMatchSnapshot();
+  });
+
+  it("should render children", () => {
+    const { container } = render(<Html>{"<b>Hello</b> World"}</Html>);
+    expect(container.firstChild.innerHTML).toBe("<b>Hello</b> World");
```
douglasduteil: This is already tested by the snapshot above

packages/code-du-travail-frontend/src/common/__tests__/Html.test.js
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  it("should render with className", () => {
+    const { container } = render(
+      <Html className="toto">{"<b>Hello</b> World"}</Html>
+    );
+    expect(container.firstChild.getAttribute("class")).toMatch(/toto/i);
```
douglasduteil: This is testing all the above assertion if you snapshot the container

packages/code-du-travail-frontend/src/search/Answer.js
```diff
@@ -73,7 +72,6 @@ class Answer extends React.Component {
               {intro}
               <Html>{html}</Html>
               {children}
-              {date && <DateContenu value={date} />}
```
douglasduteil, line 76: Seems like removing the `DateContenu` is not related to the test branch

douglasduteil (changes requested):
- Why using "react-testing-library" other "react-test-renderer"
  \cc @revolunet 
- Some "request changes" might be better done in another PR

packages/code-du-travail-frontend/src/search/Answer.js
lionelB: It is. The goal was to have `/commons` tested. DateContenu is not usefull anymore since date was available in header, that's I remove it

douglasduteil: Alright

packages/code-du-travail-frontend/src/common/FeedbackForm.js
```diff
@@ -49,62 +55,55 @@ class FeedbackForm extends React.Component {
...
+      motif,
+      message,
+      email,
+      userAgent: typeof navigator !== "undefined" && navigator.userAgent,
+      subject: question
+    };
+    this.props
+      .onSubmit(data)
+      .then(data => {
+        if (data.success) {
```
douglasduteil: Isn't it redundant ?
`then` callbacks should always be successful no ?

packages/code-du-travail-frontend/src/common/FeedbackForm.js
```diff
@@ -49,62 +55,55 @@ class FeedbackForm extends React.Component {
...
+          this.setState({
+            status: "sent",
+            email: "",
+            message: "",
+            motif: motifLabels[0]
           });
-      }
-    );
+        } else if (data.error) {
```
douglasduteil: should ne the content of the `catch(e => ...)`

douglasduteil (changes requested):

lionelB:
@douglasduteil I wrap the inner API stuff (success / error) into feedback service, I can't find detailed ressource about formspree api response. I believe @revolunet do it like that on purpose.

what do you think ?

douglasduteil (approved):

douglasduteil:
That's way better no ?

lionelB:
yes !

PR #313, opened 2018-11-26
fix(frontend): return mock of return data on outils fetch error

douglasduteil:

### DIFF
--- a/packages/code-du-travail-frontend/pages/outils.js
+++ b/packages/code-du-travail-frontend/pages/outils.js
@@ -52,7 +52,7 @@ class Outils extends React.Component {
       })
       .catch(e => {
         res.statusCode = 404;
-        throw e;
+        return { data: { _source: {} } };
       });
   }
   render() {


lionelB (changes requested):
You should add a comment of  why you choose to fail silently rather than handle the error

douglasduteil:
> why you choose to fail silently rather than handle the error

This is handling the error.
If you `throw` the code stops and the user can't see a thing.
It's unknown to me if we catch all errors in the browser to a logstash or sth
by returning a similar structure with no data it's activating the "Big Error" flow [giving the user a not found response](
https://github.com/SocialGouv/code-du-travail-numerique/pull/313/files#diff-4fd0387f53cc8b097c46a03246a8f7ebR37)

douglasduteil:
I actually changed that to test my PR with the Alert code ;)

lionelB:
> If you throw the code stops and the user can't see a thing.
In fact, throwing the error shown the error page

douglasduteil:
> In fact, throwing the error shown the error page

Alright. Is it the 500 error page ? Is it better than the error in the context ?!
Tested with https://codedutravail.num.social.gouv.fr/outils/sdf

lionelB:
well is is, but it should be 404 :)
ok to use the in context error

Le lun. 26 nov. 2018 à 14:58, Douglas Duteil <notifications@github.com> a
écrit :

douglasduteil:
> well is is, but it should be 404 :)

Should I still add a `console.error(e)` ?
How do we log error from user land back to us ?

lionelB:
@douglasduteil  we should add sentry to log error. 
I propose to use console.error for now, and create a PR to add sentry client and get rid of all console.error. what do you think ?

lionelB (approved):

PR #315, opened 2018-11-26
WIP refacto(ui): introduce typescript at the UI level <3

douglasduteil:
---

This is a migration first step. To ensure safety, the current code and build is maintained. The new ts code is used when migrated.


packages/code-du-travail-frontend/pages/outils.js
```diff
@@ -3,7 +3,8 @@ import { withRouter } from "next/router";
 import getConfig from "next/config";
 import Head from "next/head";
 import fetch from "isomorphic-unfetch";
-import { Container, Alert } from "@cdt/ui";
+import { Container } from "@cdt/ui";
+import { Alert } from "@cdt/ui/esm";
```
lionelB: why introduce another folder since we have src, lib, es. 
what the difference with es ?

packages/code-du-travail-ui/src/utils/flavors.ts
```diff
@@ -0,0 +1,95 @@
...
+      ...(flavors.includes(key) && Boolean(props[key]) === false
+        ? {}
+        : { [key]: String(props[key]) })
+    }),
+    {}
+  );
+
+//
+
+export interface FlavourProps {
```
lionelB, line 69:
```
type Success = "success"
...
type FlavourProps = Primary | Secondary | Warning | Info | Success 
```
because `<Button primary secondary>hello</Button>` does'nt make sense.

packages/code-du-travail-ui/package.json
```diff
@@ -11,29 +11,35 @@
     "umd"
   ],
   "scripts": {
-    "build": "nwb build-react-component --no-demo ",
-    "clean": "nwb clean-module && nwb clean-demo",
-    "prepublishOnly": "npm run build && npm run gh-pages",
-    "gh-pages": "npm run docz:build && gh-pages -d ./doc-build",
+    "build": "run-p -l build:*",
```
lionelB: why not simply use tsc directly ?
```
"build": "tsc --module es2015 --outDir esm",
```

packages/code-du-travail-ui/package.json
```diff
@@ -11,29 +11,35 @@
...
+    "build:esm": "tsc --module es2015 --outDir esm",
     "start": "docz dev",
-    "test": "nwb test-react",
-    "test:coverage": "nwb test-react --coverage",
-    "test:watch": "nwb test-react --server",
+    "watch": "npm-run-all \"build:esm --watch --preserveWatchOutput\"",
```
lionelB: same as build

packages/code-du-travail-ui/__tests__/Alert.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+import React from "react";
+import { create } from "react-test-renderer";
+import { Alert } from "src/components/Alert";
+
+test("<Alert>Alert !</Alert> should render an Alert component", () => {
+  const component = create(<Alert>Alert !</Alert>);
+  expect(component.toJSON()).toMatchSnapshot();
+});
+
+test("<Alert primary>Primary</Alert> should have 'alert__primary' class", () => {
```
lionelB, line 12: test label is misleading since we make no assertion on className but on snapshot content, 
if classnames change, snapshot can be updated automatically, thus relying on className force us to update all test name.

packages/code-du-travail-ui/src/utils/flavors.ts
douglasduteil: True :+1:

packages/code-du-travail-ui/package.json
douglasduteil: because I forgot to reintroduce the "old" build system for the js files :sweat_smile:

packages/code-du-travail-ui/__tests__/Alert.spec.tsx
douglasduteil: ?


> test label is misleading since we make no assertion on className but on snapshot content,

The assertion is made on everything inside a snapshot no ?

> if classnames change, snapshot can be updated automatically, thus relying on className force us to update all test name.

It's not because you can update all snapshot in one go that you should.
Snapshot are test content too. they can/should be edited by hand when necessary 

---

How would you test this ?

lionelB: Sorry my comment was misleading
I think we should use  
```
 test("<Alert primary>Primary</Alert> should render", () => { 
   expect(component.toJSON()).toMatchSnapshot();
})
```
rather than use implementation detail (eg class name we use for alert

packages/code-du-travail-ui/src/Section.js
```diff
@@ -15,7 +15,7 @@ const Section = ({ light, dark, style, children }) => {
   return (
     <section className={containerClassName} style={style}>
       <div className="container">
-        <div className={innerContainerClassName}>{children}</div>
+     <div className={innerContainerClassName}>{children}</div>
```
lionelB, line 18: nitpick : missing indentation

packages/code-du-travail-ui/.gitignore
```diff
@@ -2,6 +2,7 @@
 /coverage
 /demo/dist
 /es
+/esm
```
lionelB: not used

packages/code-du-travail-ui/package.json
```diff
@@ -11,34 +11,42 @@
...
-    "lint": "eslint src/* tests/*",
+    "lint": "tslint --project tsconfig.json -t stylish",
     "precommit": "lint-staged",
     "prepublishOnly": "npm run build && npm run gh-pages",
     "gh-pages": "npm run docz:build && gh-pages -d ./doc-build",
     "start": "docz dev",
-    "test": "nwb test-react",
-    "test:coverage": "nwb test-react --coverage",
-    "test:watch": "nwb test-react --server",
+    "watch": "npm-run-all \"build --watch --preserveWatchOutput\"",
```
lionelB, line 21: why not use `npm run build -- --watch --preserveWatchOutput`

douglasduteil:
To old now

PR #332, opened 2018-11-28
test(frontend): add global mock for next/config

lionelB:
move 'next/config' mock into jest setup so we don't have race condiftions between import and mock


douglasduteil (approved):
Thanks

PR #349, opened 2018-12-03
refactor suggest

lionelB:
refactor suggester to a controlled component
add tests

### DIFF
--- a/packages/code-du-travail-frontend/src/search/Search.js
+++ b/packages/code-du-travail-frontend/src/search/Search.js
@@ -218,8 +218,8 @@ class Search extends React.Component {
     const route = getRouteBySource(suggestion._source.source);
     const anchor = suggestion._source.anchor
       ? suggestion._source.anchor.slice(1)
-      : "";
-
+      : undefined;
+    console.log(suggestion._source.anchor);
     Router.pushRoute(
       route,
       { q: query, search: 0, slug: suggestion._source.slug },
@@ -234,8 +234,9 @@ class Search extends React.Component {
   onSearch = ({ value }) => {
     const { excludeSources } = this.state;
     fetchResultsSuggest(value, excludeSources).then(results => {
-      this.setState({ suggestions: results.hits.hits });
-      this.props.onResults(results);
+      this.setState({ suggestions: results.hits.hits }, () => {
+        this.props.onResults(results);
+      });
     });
   };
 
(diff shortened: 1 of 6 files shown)


packages/code-du-travail-frontend/src/search/Search.js
```diff
@@ -208,104 +212,98 @@ class Search extends React.Component {
...
-    if (this.props.onResults && results && results.hits.total > 0) {
-      this.props.onResults(results.hits.hits);
-    }
+
+  onSelect = suggestion => {
+    const { query } = this.state;
+    const route = getRouteBySource(suggestion._source.source);
+    const anchor = suggestion._source.anchor
+      ? suggestion._source.anchor.slice(1)
+      : "";
```
revolunet: shouldnt it be undefined here ?

packages/code-du-travail-frontend/src/search/Search.js
```diff
@@ -208,104 +212,98 @@ class Search extends React.Component {
...
+
+  onClear = () => {
+    this.setState({ suggestions: [] });
+  };
+
+  onSearch = ({ value }) => {
+    const { excludeSources } = this.state;
+    fetchResultsSuggest(value, excludeSources).then(results => {
+      this.setState({ suggestions: results.hits.hits });
+      this.props.onResults(results);
```
revolunet: maybe nest the `onResults` call in the `setState` callback ?

packages/code-du-travail-frontend/src/search/__tests__/DocumentSuggester.test.js
```diff
@@ -0,0 +1,103 @@
...
+      onSearch
+    });
+
+    const input = getByPlaceholderText(/search foo/i);
+    const mockEvent = { target: { value: "test" } };
+    fireEvent.change(input, mockEvent);
+    expect(onSearch).toBeCalledWith(expect.objectContaining({ value: "test" }));
+  });
+
+  it("should call onSelect", () => {
```
revolunet: "when user clicks some option"

revolunet (approved):

revolunet:
thanks for nice refacto +  testing 👍

packages/code-du-travail-frontend/src/search/Search.js
lionelB: hum probably but I need to test :)

packages/code-du-travail-frontend/src/search/Search.js
lionelB: I see that pattern already but I don't know why ? this defer the callback ?

revolunet: yes, this ensure the order is respected by triggering the callback only when state is effectively updated.

PR #361, opened 2018-12-07
fix(api): smaller api response on search

lionelB:
remove some field from the response returned by elastic search (mostly index related fields)


revolunet (approved):

revolunet:
bien vu

PR #396, opened 2018-12-19
chore(monorepo): run parallel prepush task before push

### LINKED ISSUE #384: add pre-push hook to lint code
It would be nice to have a git-hook to prevent push when code is not lint correctly

douglasduteil:
---

Closes #384


douglasduteil:
I added tests too.
On my computer the a push now takes ~50s

PR #398, opened 2018-12-19
test(frontend): silent  warning

douglasduteil:
---

We repeatedly have this waring in our tests :

```
    console.warn ../../node_modules/@reach/utils/index.js:16
      @reach/dialog styles not found. If you are using a bundler like webpack or parcel include this in the entry file of your app before any of your own styles:

          import "@reach/dialog/styles.css";

        Otherwise you'll need to include them some other way:

          <link rel="stylesheet" type="text/css" href="node_modules/@reach/dialog/styles.css" />

        For more information visit https://ui.reach.tech/styling.
```

This is due to an internal `@reach/utils` styles check on CSS Custom Properties.
As @lionelB stated [5b83354@/packages/code-du-travail-frontend/test/jest.setup.js](https://github.com/SocialGouv/code-du-travail-numerique/blob/5b83354ebe05369aee67eb1bbef94d9d240c6e53/packages/code-du-travail-frontend/test/jest.setup.js),
jsdom doesn't support CSS Custom Properties for now @see https://github.com/jsdom/jsdom/issues/1895
So we hack it by mocking the check itself 💩

### DIFF
--- a/packages/code-du-travail-frontend/test/jest.setup.js
+++ b/packages/code-du-travail-frontend/test/jest.setup.js
@@ -11,6 +11,15 @@ jest.mock("react-piwik", () => ({
   push: jest.fn()
 }));
 
-// trick to prevent @reach-modal warning if styles are not imported
+// HACK(lionelB): trick to prevent @reach-modal warning if styles are not imported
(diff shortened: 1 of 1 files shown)


lionelB (approved):
cheering_minions

PR #402, opened 2018-12-19
feat(api): add version endpoint

### LINKED ISSUE #391: fair remonter le numéro de version dans lerna.json dans API et frontend
- [ ] pour api: prevoir une route /version 
- [x] pour le front: prévoir un easteregg  dans un title sur un élément de la Home

douglasduteil:
---

Closes #391


lionelB (approved):
👍

PR #403, opened 2018-12-20
ci: change flag format

douglasduteil:
Following https://docs.codecov.io/docs#section-flag-creation

Flags must be lowercase, alphanumeric, and not exceed 45 characters
Only lowercase, alphanumeric values are accepted. `^[a-z0-9_]{1,45}$`


lionelB (approved):
👍

PR #436, opened 2019-01-17
Lionelb/fix elastic mapping

lionelB:
this PR
- add kibana in dev
- optimize mapping for text field


docker-compose.override.prod.yml
```diff
@@ -37,11 +37,3 @@ services:
       - setup.kibana.host=kibana:5601
       - output.elasticsearch.hosts=elasticsearch:9200
       - output.elasticsearch.enabled=true
-
```
douglasduteil, line 40: No Kibana in production ?

douglasduteil:
Why do you need kibana in dev ?
Are you making some cool dashboards with it ?
We might want to share the dev kibana config if yes.

lionelB:
@douglasduteil kibana is needed in dev in order to use the developer console. it's like graphiql, you can write request and see the results.

docker-compose.override.prod.yml
lionelB: my bad, I didn't realize I removed it from prod. we just need it in dev too :)

douglasduteil:
You have a kibana config in dev then right ?
With the default index to use at least

lionelB:
Initially, kibana was only use in prod and the ES consultant told us it could be really handy to have it in dev too, So i removed it from `docker-compose.override.prod.yml`and move it to `docker-compose.override.yml`

douglasduteil:
> So i removed it from `docker-compose.override.prod.yml`and move it to `docker-compose.override.yml`

Yes that's cool but, now that you are using it in dev, you should have a valide config for kibana that works in dev and prod (in docker) right ?

lionelB:
why do you think it's not the case ?

douglasduteil:
> why do you think it's not the case ?

I found no proof of it being the case
Can you show it to me ?

douglasduteil (approved):

PR #443, opened 2019-01-21
ci: pass api test env inside package.json

lionelB:
move env variable to package.json, since test should always use dedicated index


douglasduteil (approved):
OK

PR #444, opened 2019-01-21
api: define api docs dir through env

lionelB:
configurable docs directory for api


lionelB:
closing in favor of  #445

PR #446, opened 2019-01-21
ci: update build for pull requeset from forks

douglasduteil:


douglasduteil:
Bouya 
@revolunet @lionelB on a un script de PR qui build en 3 minutes !

PR #450, opened 2019-01-21
Lionelb/refacto api search endpoint

lionelB:
j'aimerai bien votre avis pour la suite du refacto, 

A terme, je pensais supprimé api.js et dataSources/ et avoir pour chaque route, un dossier avec un fichier pour la route et un fichier pour la query

qu'est ce que vous en dites ?


packages/code-du-travail-api/src/server/routes/__tests__/search.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    "idcc",
+    "kali",
+    "themes",
+    "fiches_ministere_travail",
+    "fiches_service_public"
+  ];
+  const response = await request(app.callback()).get(
+    `${API_BASE_URL}${routeName}?q=démission&excludeSources=${excludeSources.join(
+      ","
+    )}`
```
revolunet: what about `source=faq,themes` `source=-faq` ?

packages/code-du-travail-api/src/server/routes/search/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+  let excludeSources = (ctx.request.query.excludeSources || "").split(",");
+
+  const body = getSearchBody({ query, size, excludeSources });
+
+  ctx.body = await elasticsearchClient.search({ index, body });
+});
+
+module.exports = router;
+module.exports.routeName = `/search`;
```
revolunet: use defined `routeName`?

douglasduteil:
Un grand OUI à la séparation des routes par fichier

revolunet (commented):
+1 pour la separation des routes

packages/code-du-travail-api/src/server/routes/search/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * Return documents matching the given query.
+ *
+ * @example
+ * http://localhost:1337/api/v1/search?q=incapacité%20travail
+ *
+ * @param {string} querystring.q A `q` querystring param containing the query to process.
+ * @returns {Object} Results.
+ */
+router.get(routeName, async ctx => {
+  let query = ctx.request.query.q;
```
douglasduteil: Looks more like a const than a let

packages/code-du-travail-api/src/server/routes/search/index.js
```diff
@@ -0,0 +1,33 @@
...
+  const size = Math.min(ctx.request.query.size || 10, 100);
+
+  let excludeSources = (ctx.request.query.excludeSources || "").split(",");
```
douglasduteil: Looka more like a const than a let

packages/code-du-travail-api/src/server/routes/__tests__/search.spec.js
lionelB: good idea, I like it a lot :)  
I have also planned to update that part to use elasticsearch `filters` rather that `must_not` query

packages/code-du-travail-api/src/server/routes/search/index.js
lionelB: not sure about that, it was to re-use the routName durring test, but hardcoded string is not a big deal :)

packages/code-du-travail-api/src/server/routes/__tests__/search.spec.js
```diff
@@ -0,0 +1,36 @@
...
+const API_BASE_URL = require("../api").BASE_URL;
+const routes = require("../search");
+const { routeName } = routes;
+
+const app = new Koa();
+app.use(routes.routes());
+
+test("return search results for demission", async () => {
+  const response = await request(app.callback()).get(
+    `${API_BASE_URL}${routeName}?q=démission`
```
douglasduteil: I think I was wrong reusing the `API_BASE_URL` in the first place. The E2E tests shouldn't depend on the implementation. They should be the specifications. 
`/api/v1/search` should be apparent IMO

packages/code-du-travail-data/dataset/synonyms/package.json
```diff
@@ -5,6 +5,6 @@
   "main": "index.js",
   "license": "MIT",
   "scripts": {
-    "start": "node index.js > ../thesaurus/TESS_synonyms.json"
+    "start": "node index.js > ../synonyms.json"
```
douglasduteil: Can be part of another PR focused on the dataset changes

douglasduteil (changes requested):

douglasduteil:
Do we really need to test against data that big ?
I feel like the "fake" should be scoped by the tests them self or they will end up having the would x times the database in it...

packages/code-du-travail-api/src/server/routes/__tests__/search.spec.js
lionelB: 👍

packages/code-du-travail-api/src/server/routes/search/index.js
lionelB: I though there is a linter rule that handle that kin of nitpick

douglasduteil (approved):

packages/code-du-travail-api/src/server/routes/search/index.js
douglasduteil: I would gladly add a "no let" rule.
I think it would be better then me always commenting on every "let" I see :stuck_out_tongue: 
Still it might lift a lot of them to "fix" :smirk:

lionelB: we can do that on API, there no other error about `let`/ `const`

PR #455, opened 2019-01-22
refactor(api): move getSingleItems to separate files

lionelB:


packages/code-du-travail-api/src/server/routes/items/index.js
```diff
@@ -0,0 +1,75 @@
...
+  const relatedItems =
+    ctx.params.source === "faq" ? await searchItemsFromTheme(item) : {};
+
+  ctx.body = {
+    ...item,
+    relatedItems
+  };
+});
+
+router.get(`${routeName}/:id`, async ctx => {
```
revolunet, line 52: maybe add some jsdoc. is this stille used ?

lionelB: not used for now, but could be one possible fix for our uniq slug problem (move from slug to id in url)

revolunet (approved):

packages/code-du-travail-api/src/server/routes/items/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+  if (results.aggregations.bySource.buckets.length === 0) {
+    return {};
+  }
+
+  const { buckets } = results.aggregations.bySource;
+
+  return buckets.reduce((state, bucket) => {
+    if (bucket.doc_count > 0) {
+      state[bucket.key] = bucket.bySource.hits.hits;
```
douglasduteil: could be :

```js
return bucket.doc_count > 0 ? { ...state, [bucket.key]: bucket.bySource.hits.hits } : state
```

:heart_eyes:

douglasduteil (approved):
C'est de toute beauté :heart:

PR #457, opened 2019-01-22
Lionelb/refactor api annuaire

lionelB:
refactor api to separate files 
I update `packages.json` to avoid jest error/warning on `*.mock.js` since they are not test files.
I mock `node-fetch` module to return the location of loriol (near an valence  annuaire document)
I've also updated the elastic search data to have two annuaire document that are farther to only retrieve one document depending on location


revolunet (approved):

packages/code-du-travail-api/src/server/routes/__tests__/annaire.spec.js
```diff
@@ -0,0 +1,29 @@
...
+});
+
+test("should return all the records around loriol", async () => {
+  fetch.mockResolvedValue({ json: () => addresses.payload });
+
+  const response = await request(app.callback()).get(
+    `/api/v1/annuaire/search?q=${addresses.query}`
+  );
+  expect(response.status).toBe(200);
+  expect(response.body.hits).toMatchSnapshot();
```
douglasduteil, line 30: Missing `expect` on fetch.
We should ensure that it's called. 
We don't need to mock it if not

packages/code-du-travail-api/src/server/routes/__tests__/api-adresse.mock.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
```
douglasduteil: It looks like a `json` file to me no ?

packages/code-du-travail-api/src/server/routes/annuaire/index.js
```diff
@@ -0,0 +1,66 @@
...
+const elasticsearchClient = require("../../conf/elasticsearch.js");
+const getAnnuaireBody = require("./annuaire.elastic");
+
+const router = new Router({ prefix: API_BASE_URL });
+const index = process.env.ELASTICSEARCH_ANNUAIRE_INDEX || "cdtn_annuaire";
+const routeName = "/annuaire";
+
+// todo @lionelb move to constant file
+const defaultMaxSearchDistance = "30km";
+const adresseApiEndPoint = "https://api-adresse.data.gouv.fr/search";
```
douglasduteil, line 14: IMO those const looks like configurations.
We might consider extracting them from the code.

packages/code-du-travail-api/src/server/routes/annuaire/index.js
```diff
@@ -0,0 +1,66 @@
...
+ * @example
+ * http://localhost:1337/api/v1/annuaire/search?coord=lon:lat
+ * http://localhost:1337/api/v1/annuaire/search?q=address to search
+ *
+ * @param {string} :base coord or address for geo search
+ * @returns {Object} Result.
+ */
+router.get(`${routeName}/search`, async ctx => {
+  const { coord, q } = ctx.request.query;
+  const results = searchFromCoord(coord) || (await searchFromQuery(q));
```
douglasduteil: :heart: that 
Are the `()` parenthesizes really required ?

packages/code-du-travail-api/src/server/routes/annuaire/index.js
```diff
@@ -0,0 +1,66 @@
...
+    ctx.body = await elasticsearchClient.search({
+      index,
+      body
+    });
+  } else {
+    ctx.body = { total: 0, hits: { hits: [] } };
+  }
+});
+
+function searchFromCoord(coord) {
```
douglasduteil, line 46: If we want to go full functional `searchFromQuery` should be curried to take the `adresseApiEndPoint` as first arguments

```js
async function searchFromQuery(adresseApiEndPoint){ 
  return query => {
    if (!query) {
      return false;
    }
  // [...]
```

 It's looking very good already :)

douglasduteil (changes requested):
Love this coding style :100:

packages/code-du-travail-api/src/server/routes/annuaire/index.js
lionelB: yup, move it to a separate file ie `route.config.json` // `api.config.{json|js}`

packages/code-du-travail-api/src/server/routes/annuaire/index.js
lionelB: > Are the `()` parenthesizes really required ?
prettier :)

packages/code-du-travail-api/src/server/routes/annuaire/index.js
lionelB: currying is not needed for now, so I only pass endpoint as a parameter so we could curry it later if we need it

lionelB:
@douglasduteil updated !

packages/code-du-travail-api/package.json
```diff
@@ -52,8 +52,13 @@
     "supertest": "^3.3.0"
   },
   "jest": {
+    "testPathIgnorePatterns": [
+      "node_modules",
+      ".mock.js$"
+    ],
     "collectCoverageFrom": [
       "src/**/*.js",
+      "!src/**/*.mock.js",
```
douglasduteil: Still required now that mocks are json files ?

lionelB: no no no I update

lionelB:
@douglasduteil 🆙

douglasduteil (approved):
Super

PR #459, opened 2019-01-23
chore(frontend): remove next-plugin-transpile-modules

lionelB:
next-plugin-transform-module broke reload from `packages/*` files (eg updatating css need a next restart :/ )
I create a local copy of pDebounce so it will be transpile

### DIFF
--- a/packages/code-du-travail-frontend/next.config.js
+++ b/packages/code-du-travail-frontend/next.config.js
@@ -1,3 +1,4 @@
+const BabelEnginePlugin = require("babel-engine-plugin");
 const withCSS = require("@zeit/next-css");
 module.exports = withCSS({
   // https://github.com/zeit/next.js/#disabling-file-system-routing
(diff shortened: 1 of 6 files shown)


douglasduteil:
This is bad...
We should avoid copy pasting node_modules in our code.
Are they other solutions ?

revolunet:
Upstream pull request ?

douglasduteil:
Upstream => next-plugin-transform-module ?
Ou 
Upstream => pDebounce ?

lionelB:
I try to make next-plugin-transpile-module works without success so bafore wasting to much time, I prefer fix the problem we actually have (css update need next to restart) and figure a better solution

revolunet:
sent a PR to upstream project and... got rejected :) https://github.com/sindresorhus/p-debounce/pull/4

there are some alternatives like https://github.com/bjoerge/debounce-promise

lionelB:
For the record, babel-engine-plugin not work for ie, so I leave it with code duo. I think p-debounce should not evolve that much

lionelB:
@douglasduteil  est ce qu'on peut merger ?

revolunet (approved):

PR #460, opened 2019-01-23
Lionelb/refactor clean api

lionelB:
last bits of cleaning


douglasduteil (approved):
- Can you move all 

	```js
	const router = new Router({ prefix: API_BASE_URL });
	```
	
  So it's not mixed with the `require` statements ?

- What do you think about merging the prefix and the `routeName`

	```js
	const router = new Router({ prefix: `${API_BASE_URL}/items` });
	router.get('/', [...])
	```

lionelB:
> * What do you think about merging the prefix and the `routeName`
>   ```js
>    const router = new Router({ prefix: `${API_BASE_URL}/items` });
>    router.get('/', [...])
>   ```

I found it more readable and easier to grasp which endpoint it refer when using the actual notation

```diff
+ const router = new Router({ prefix: `${API_BASE_URL}/idcc` })
+ router.get("", async ctx => {
- const router = new Router({ prefix: `${API_BASE_URL}` })
- router.get("/idcc", async ctx => {

```

douglasduteil:
Alright :+1:

PR #466, opened 2019-01-24
fix(feedback.service): use a new formspree endpoint to handle prod+dev

revolunet:
this form should work for both urls


lionelB (approved):
cool! merci

PR #473, opened 2019-01-29
fix(frontend): scroll up after url change

### LINKED ISSUE #467: Annuaire: scrollTop on adress selection
When user selects an annuaire suggestion in the pge footer, the results are displayed, but scroll is not reset so user is not aware of the results update. we should add something like `scrollTop=0`

lionelB:
fix #467


revolunet (approved):
😍

PR #475, opened 2019-01-29
Lionelb/ci update lint rule

lionelB:
just a small addition to add `prefer-const` rule 

And then things gone crazy as I wanted to fixed every rules (damn `jsx-a11y/anchor-is-valid`) but I managed to temper and only remove 1 rules :D


revolunet (approved):

douglasduteil:
... Might want to be added to https://github.com/SocialGouv/linters/blob/master/packages/eslint-config-recommended/index.js instead...

PR #479, opened 2019-01-29
fix(frontend): add fiche url when submiting feedback

lionelB:
ajout de l'url de la page consultée fors d'une remontée utilisateur via le formulaire de feedback
fixe #316

### DIFF
--- a/packages/code-du-travail-frontend/src/common/Answer.js
+++ b/packages/code-du-travail-frontend/src/common/Answer.js
@@ -72,6 +72,7 @@ class Answer extends React.Component {
               isOpen={this.state.modalVisible}
               closeModal={this.closeModal}
               query={router.query.q || title}
+              fiche={`${router.route}/${router.query.slug}`}
             />
             <Article
               title={title}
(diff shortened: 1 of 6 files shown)


packages/code-du-travail-frontend/src/common/Answer.js
```diff
@@ -72,6 +72,7 @@ class Answer extends React.Component {
               isOpen={this.state.modalVisible}
               closeModal={this.closeModal}
               query={router.query.q || title}
+              fiche={`${router.route}/${router.query.slug}`}
```
revolunet: ce serait pas plutôt `url` que `fiche` ?

revolunet: `Router.asPath` (string) should work too

revolunet: wondering if `document.location.href`, at runtime, wouldnt be better so we also get hostname :)

lionelB: why not =)

douglasduteil (approved):

packages/code-du-travail-frontend/src/common/Answer.js
lionelB: I've updated the component use `document.location.href` while submiting for using JS and component can also receive an url for server rendered case or when there is no javascript

revolunet (approved):

PR #506, opened 2019-02-06
chore: update lint config

lionelB:
cette PR est un découpage de la grosse PR #501
- mise à jour de la config `@socialgouv/eslint-config-react`


douglasduteil (approved):
LGTM

PS: Does renaming `Cards` has an effect of the frontend app ?

lionelB:
> PS: Does renaming `Cards` has an effect of the frontend app ?

No since it was the default export

note: I generally prefer to use a name export so that in case of refactoring, all occurences are renamed accordingly.

PR #514, opened 2019-02-08
Lionelb/update layout for search

lionelB:
- split layout for Home and Pages Components 
- hide disclaimer text once a search had been done 
image


packages/code-du-travail-frontend/src/search/Search.js
```diff
@@ -101,6 +101,7 @@ class Search extends React.Component {
         : this.props.router.query.q
     });
 
+    console.log("---", "routeChange");
```
douglasduteil: trailling console.log

douglasduteil (changes requested):

douglasduteil (approved):

PR #523, opened 2019-02-12
ci: remove long and rarely breaking manual build stage

douglasduteil:
---

Speed up `Branch` build by `~2.9x` by removing constant `~2m` build step

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,12 +38,19 @@ jobs:
         - *docker_login
         - docker push $DOCKER_IMAGE
 
-    - <<: *main_node_stage
+    - &main_node_stage
+      stage: Prepare
       if: type = pull_request
       name: Pull Request Stage
+      language: node_js
+      node_js: 10
+      cache: yarn
       env:
         - ELASTICSEARCH_LOG_LEVEL=info
-      install: yarn
+      before_install:
+        - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.13.0
+        - export PATH="$HOME/.yarn/bin:$PATH"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,23 +40,17 @@ jobs:
 
     - &main_node_stage
       stage: Prepare
-      if: NOT env(RELEASE) AND NOT type = pull_request
+      if: type = pull_request
```
lionelB, line 43: Could it be a problem, that will not trigger on tag (I think it was the case before)  ?
On the `main_node_state` for PR we also do a `yarn build --stream` does it use the yarn cache ?

douglasduteil: > Could it be a problem, that will not trigger on tag (I think it was the case before) ?

It was the case when the stage name was "Monorepo build script". The goal was to have a separate `build` task in parallel of the `build` task running inside Docker `Build and deploy current commit docker image` stage. As that task was always successful (we rarely break the build), I proposing to remove it.

 

> On the `main_node_state` for PR we also do a `yarn build --stream` does it use the yarn cache ?

Yes it does. As PRs doesn't have the right to decode our security keys, we need special PR stages like this one.
You can see it in action here : https://travis-ci.com/SocialGouv/code-du-travail-numerique/builds/100641864  
And compare it with the Branch build : https://travis-ci.com/SocialGouv/code-du-travail-numerique/builds/100641855

lionelB: ok ^_^ about the tag, I remember we had a problem when we created a release, since the deploy was only running for PR and not branch (in case of a tag)

lionelB (approved):

PR #539, opened 2019-02-20
refactor(monorepo): minor setup-env script changes

UnbearableBear:


douglasduteil (approved):
Why not. You could have called it `copyAsync` etc..

PR #582, opened 2019-03-08
doc: enforce NodeJS 10.15 and improve local tests doc

adipasquale:


README.md
```diff
@@ -76,6 +78,9 @@ $ yarn workspace @cdt/frontend build
 ### Test
 
 ```sh
+# Prepare ES indexes for the test environment
```
lionelB: can you add that it will create a dedicated index for documents (`cdtn_document_test`  and `cdtn_annuaire_test`)
and the fact that this command is only needed the first time

adipasquale: ok! but it's needed every time you make changes to the indexing code no? (changes that impact tests, that is)

lionelB: that's right !  it's also needed if you remove the docker volume where the ES indexes are saved

lionelB (approved):
👍

PR #618, opened 2019-03-21
chore: run pre push hook scripts in parallel

douglasduteil:
---

Because we can make the CPU go :100: during 1min ! :fire: :desktop_computer: :fire:

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   ],
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


lionelB (approved):
`git push --no-verify`  :trollface:

douglasduteil:
`rm -rf node_modules/husky`  :trollface:

PR #676, opened 2019-04-04
ci(travis): disable git --depth

douglasduteil:
---

Try to have all the commits for the release Change Log


lionelB (approved):
Do you plan to do another PR to check if release note generation work as expected ?
We will do a release these days, It could be a good test :)

PR #713, opened 2019-04-11
pref(docker): reorder copies in master image for better caching

douglasduteil:
---

By order of "more likely to change" the frontend, the api, etc... are changing more often than the dataset.
Putting them after will optimise the native docker build cache of the image


revolunet (approved):

lionelB (approved):
looks good! have you some documentation about the perf optimisation ?

douglasduteil:
https://docs.semaphoreci.com/article/81-docker-layer-caching#the-copy-command

PR #729, opened 2019-04-15
First UI for Convention Collective contents using tabs and a sidebar in the summary

adipasquale:
this is the first part of the issue #750 that aims at displaying Conventions contents. The main missing bits are the full-text search and the automatic scrolling.

This PR focuses on the `kali` page:
- This page now fetches additional data from the `dila2sql` API on https://api.dila2sql.num.social.gouv.fr (conteneurs and textes). e.g. https://api.dila2sql.num.social.gouv.fr/v1/base/KALI/conteneur/KALICONT000005635871
- There is a top expandable panel with generic information to understand what's a convention. (the text is temporary and being redacted by @virginielastisneres but we can deploy as is IMHO). 
- We add tabbed panels (thanks @UnbearableBear). 
- There is a sidebar that displays the table of contents and allows expanding the top level of a text
- The first tab displays metadatas about the Convention as a whole. 
- The next tabs let you see the texts that compose the Convention one by one.
- I've reindexed the `kali` dataset (the list of Conventions indexed in ES) to only include the active ones. The other ones may be more tricky to display and are out of scope so far in any case. 
- I've written two very simple snapshot tests that only check that the `Convention` and the `ConventionTexte` components display. I don't really think we should write more thorough tests with the interactions yet as the UX is very probably going to change soon. 

screenshots:

Screen Shot 2019-05-14 at 15 29 17

Screen Shot 2019-05-14 at 15 29 34

Screen Shot 2019-05-14 at 15 29 41

Notes: 
- I had implemented a first version of the sidebar auto-scroll but I removed it as it was not satisfying. this commit cf90e317800914b515b02cea916254ddad8af1c5 could be used to iterate later.
- The 15k lines mostly come from the snapshot of the `ConventionTexte`


lionelB:
> please don't review

ahah :)

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
```diff
@@ -0,0 +1,90 @@
...
+import { Section, Wrapper } from "@cdt/ui";
+import styled from "styled-components";
+
+const storedExpandedKey = "isConventionExplainerExpanded";
+
+const ConventionExplainer = () => {
+  const [expanded, setExpanded] = useState(true);
+
+  useEffect(() => {
+    const expanded = sessionStorage.getItem(storedExpandedKey) == "true";
```
lionelB: why not use `JSON.parse()` / `JSON.stringify(!expanded)` ?

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
```diff
@@ -0,0 +1,90 @@
...
+    setExpanded(expanded);
+  });
+
+  return (
+    <Section>
+      <Wrapper variant="dark">
+        <ExpandWrapper>
```
lionelB: Why did you not  use an `<Accordion />` here ?

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
```diff
@@ -0,0 +1,90 @@
...
+            <p>
+              Pour pouvoir s&apos;appliquer à vous, la convention collective de
+              branche doit être applicable à votre entreprise.
+              <br />
+              Lorsque votre entreprise entre dans le champ professionnel et
+              géographique d&apos;une convention collective de branche, il
+              existe deux situations dans lesquelles votre entreprise a
+              l&apos;obligation de l&apos;appliquer :
+              <br />
+              - soit une convention collective de branche a été étendue par le
```
lionelB: whu not use  a `<ul>` `<li>` markup here  ?

packages/code-du-travail-frontend/src/conventions/ConventionInfos.js
```diff
@@ -0,0 +1,33 @@
...
+
+        <p>
+          <a target="_blank" rel="noopener noreferrer" href={url}>
+            Voir la convention sur Legifrance ↗
+          </a>
+        </p>
+      </React.Fragment>
+    );
+  }
+}
```
lionelB, line 32: Can you add the component propTypes here 
```js
ConventionInfos.propTypes = {
  convention: PropTypes.shape({
    url: PropTypes.string
  }).isRequired 
  conteneur: PropTypes.shape({
    num: PropTypes.number,
    date_publi: PropTypes.string 
  }).isRequired
}

File: packages/code-du-travail-frontend/src/conventions/Convention.js [hunk 5]
```diff
@@ -0,0 +1,60 @@
...
+              typeTextes={"extensions"}
+            />
+          ),
+          key: "extensions"
+        }
+      ]}
+    />
+  </React.Fragment>
+);
+
```
lionelB, line 63: Can you add the components propTypes ?

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
```diff
@@ -0,0 +1,69 @@
...
+  publicRuntimeConfig: { API_DILA2SQL_URL }
+} = getConfig();
+
+class ConventionListeTextes extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { loaded: false, textes: [], selectedTexte: null };
+  }
+
+  fetchTextes({ conteneurId, typeTextes }) {
```
lionelB: this should be in a `convention.service.js`

packages/code-du-travail-frontend/src/conventions/texte/Calipso.js
```diff
@@ -0,0 +1,20 @@
...
+      borderRadius: "2px",
+      margin: "2px",
+      fontSize: "0.8rem",
+      padding: "2px 10px"
+    }}
+  >
+    {calipso}
+  </div>
+);
+
```
lionelB, line 18: can you add the propTypes too ?

packages/code-du-travail-frontend/src/conventions/texte/Calipso.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+const Calipso = ({ calipso }) => (
+  <div
+    key={calipso}
```
lionelB: why do you need to add a key here ?

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -0,0 +1,109 @@
...
+    if (this.state.texte) {
+      return false;
+    }
+    const { id } = this.props;
+    const texte = await this.fetchTexte({ id });
+    const rootNode = this.getRootNode(texte);
+    this.setState({ texte, rootNode, loaded: true });
+  }
+
+  getRootNode(startNode) {
```
lionelB: I found the name misleading, I imagine the function go up the tree to find the root node of a node but this function  seems to go down the children

lionelB (changes requested):
@adipasquale I didn't fully finish the review but 
I found It could help if you add some sample data as comment to help understand the input of a component or at least add the propTypes

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
adipasquale: I don't see how that would be better @lionelB ? my arguments: 
- I don't see this value getting more complex than it is now (a boolean)
- I think it's easier to read as-is
- I guess it's more performant as-is

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
adipasquale: I didn't check that we had such a component 🤦‍♂ I'll change to use it

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
adipasquale: 👌 I'll update it. (this content is quite temporary so I just copy pasted)

packages/code-du-travail-frontend/src/conventions/ConventionInfos.js
adipasquale: 👍 will do

packages/code-du-travail-frontend/src/conventions/Convention.js
adipasquale: 👍 will do

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
adipasquale: good call, I'll move it 👍

packages/code-du-travail-frontend/src/conventions/texte/Calipso.js
adipasquale: this must be a mistake from a previous version, good call

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
adipasquale: I get what you're saying, could `getFirstNodeWithChildren` be better?

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
lionelB: because it's nitpick, can you just add a comment about the type coercion

> I think it's easier to read as-is

Not to me :) When I read the code I was wondering, "hey, why does it compare a string, weird ! This should be a Boolean" and then I remember that storage are string based.

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
lionelB: totally !

packages/code-du-travail-frontend/src/conventions/ConventionExplainer.js
adipasquale: actually I removed completely this localStorage based behaviour because I switched to using the `Accordion` component which doesn't support it

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import PropTypes from "prop-types";
+import ConventionTexte from "./ConventionTexte";
+import { fetchTextes } from "../common/convention.service";
+
+class ConventionListeTextes extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { loaded: false, textes: [], selectedTexte: null };
```
lionelB: For initialialize the state, I would prefer  this notation 
```
state = {
  loaded: false
  textes: []
  selectedTexte: null
}

```

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
```diff
@@ -0,0 +1,66 @@
...
+          </React.Fragment>
+        )}
+        {loaded && !selectedTexte && textes.length == 0 && (
+          <div>Aucun textes</div>
+        )}
+        {loaded && !selectedTexte && textes.length > 0 && (
+          <ul>
+            {textes.map(texte => (
+              <li key={texte.id}>
+                <a href="#" onClick={e => this.onSelectTexte(e, texte)}>
```
lionelB: can you use <Button variant="link" /> here since its not a link ?

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -0,0 +1,108 @@
...
+import { fetchTexte } from "../common/convention.service";
+import Sidebar from "./texte/Sidebar";
+import Content from "./texte/Content";
+import styled from "styled-components";
+import { theme } from "@cdt/ui/";
+
+class ConventionTexte extends React.Component {
+  constructor(props) {
+    super(props);
+    if (props.preloadedTexte) {
```
lionelB, line 12: I can't find where this props `preloadedTexte` is used ? Is it still used ?

adipasquale: it's in the tests, yes I believe it's used

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
adipasquale: a button you mean I guess ? sure thing, I'll update it

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -0,0 +1,108 @@
...
+
+  getFirstNodeWithChildren(texte) {
+    let topNode = texte;
+    while (topNode.children && topNode.children.length == 1) {
+      topNode = topNode.children[0];
+    }
+    return topNode.children ? topNode : texte;
+  }
+
+  onChangeSummaryTitleExpanded(expanded, sectionId) {
```
lionelB: you can use this notation too 
```
onChangeSummaryTitleExpanded = (sectionId, expanded) => {
}
```

so you can directly use it in render 
```
<Sidebar
              topNode={topNode}
              onSummaryTitleToggleExpanded={this.onChangeSummaryTitleExpanded}
            />
```

packages/code-du-travail-frontend/src/conventions/texte/Sidebar.js
```diff
@@ -0,0 +1,35 @@
...
+            {child.children}
+          </SummaryTitle>
+        ))}
+      </Wrapper>
+    </Section>
+  );
+};
+
+Sidebar.propTypes = {
+  topNode: PropTypes.shape({
```
lionelB: topNode could be name node

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
```diff
@@ -0,0 +1,69 @@
...
+const SummaryItem = ({ data, children, type, level, visible }) => {
+  const { titre, id } = data;
+  const borderColor = 0 + (level + 1) * 50;
+  const style = {
+    marginLeft: `5px`,
+    paddingLeft: `10px`,
+    borderLeft: `1px solid rgba(${borderColor}, ${borderColor}, ${borderColor}, 1)`
+  };
+  return (
+    <div style={style} title={`${type} ${id}, level ${level}`}>
```
lionelB: This should be un a styledComponent

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
```diff
@@ -0,0 +1,69 @@
...
+      <Title visible={visible}>
+        <TitleLink href={`#${id}`}>{titre}</TitleLink>
```
lionelB, line 11: Good point for the navigation anchor

packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
lionelB: yes it was a `<Button variant="link" />`

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
adipasquale: thanks!

lionelB (approved):
Thanks a lot and bravo !

PR #791, opened 2019-04-30
feat(api): add themes endpoint

lionelB:
add themes endpoint to api
`/api/v1/themes/41-licenciement-pour-motif-personnel`

will return 
```
{
  theme: {id: 41, title: "Licenciement pour motif personnel", slug: "41-licenciement-pour-motif-personnel" },
  breadcrumbs: [{}], // an array containing the parent themes 
  children: [{}] // an array containing the children theme
}
```

### DIFF
--- a/packages/code-du-travail-data/dataset/themes/index.js
+++ b/packages/code-du-travail-data/dataset/themes/index.js
@@ -18,18 +18,29 @@ const slugify = require("slugify");
  *
  *
  *  and will return a flat list of theme
- * {
+ * [{
  *  id: 1,
  *  label: "Recrutement et contrat de travail",
- *  slug: "recrutement-et-contrat-de-travail"
+ *  slug: "1-recrutement-et-contrat-de-travail"
  * },
  * {
  *  id: 9,
  *  label: "Recrutement",
- *  slug: "recrutement"
- *  parent: "recrutement-et-contrat-de-travail"
+ *  slug: "9-recrutement"
+ *  parent: "1-recrutement-et-contrat-de-travail"
  * },
- *
+ * {
+ *  id: 50,
+ *  label: "Méthodes de recrutement'",
+ *  slug: "50-methodes-de-recrutement'"
+ *  parent: "9-recrutement"
+ * },
+  * {
+ *  id: 51,
+ *  label: "Curriculum vitae (CV)",
+ *  slug: "51-curriculum-vitae-cv"
+ *  parent: "9-recrutement"
+ * }]
  */
 
 function flattenThemes(data) {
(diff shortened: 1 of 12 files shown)


packages/code-du-travail-data/dataset/themes/query.js
```diff
@@ -0,0 +1,26 @@
+const themes = require("./themes.json");
+const memoizee = require("memoizee");
```
UnbearableBear, line 2: I would have called it `memoize` but I don't care ^^

lionelB: on the other hand we know that's the memoize function from a 3rd party lib ;)

packages/code-du-travail-data/dataset/themes/index.js
```diff
@@ -0,0 +1,92 @@
+const cstojson = require("csvtojson");
+const slugify = require("slugify");
+
+/**
+ *
+ * flattenThemes take a csv file of themes
```
UnbearableBear, line 6: Very nice to sample some data here

packages/code-du-travail-data/dataset/themes/index.js
```diff
@@ -0,0 +1,92 @@
...
+ * }, {
+ *    niveau1: '1 - Recrutement et contrat de travail',
+ *    niveau2: '9 - Recrutement',
+ *    niveau3: '51 - Curriculum vitae (CV)',
+ *    niveau4: ''
+ * }]
+ *
+ *
+ *  and will return a flat list of theme
+ * {
```
UnbearableBear: Does the output here matches the sample above ?? I'm not sure I get the logic if so

lionelB: missing 50 and 51 in the output

UnbearableBear: Ok got it !

packages/code-du-travail-data/dataset/themes/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Extract a slug from a chained theme breadcrumbs
+ * input : 46 - Conflits au travail et contrôle de la réglementation - Sanctions disciplinaires
+ * output : 46-sanctions-disciplinaires
+ * @param {String} str complete str
+ * @returns {String} a slugify version of the id+theme
+ */
+
+function extractSlug(str) {
```
UnbearableBear, line 91: Ie this one used somewhere ?

UnbearableBear (approved):
Nice !!

packages/code-du-travail-data/dataset/themes/index.js
lionelB: yes the function is exported to be used in `cdtn-theme-{sp|mt}.js` scripts

PR #793, opened 2019-05-02
feat(ui): add toast component

UnbearableBear:
Tout beau tout chaud !
Screenshot 2019-05-02 at 12 20 46


lionelB (dismissed):
Can you clean your PR with only Toast related files ?

PR #795, opened 2019-05-02
refactor(ui): button props

UnbearableBear:
#783 step 1


packages/code-du-travail-ui/src/ToggleButton/index.js
```diff
@@ -4,24 +4,34 @@ import styled, { css } from "styled-components";
...
+// eslint-disable-next-line
+const Button = ({ pressed, kind, onClick, ...props }) => {
+  return <button aria-pressed={pressed} {...props} onClick={onClick} />;
 };
 
-ToggleButton.propTypes = {
-  pressed: PropTypes.bool,
-  onClick: PropTypes.func
+Button.propTypes = {
+  kind: PropTypes.oneOf([
```
lionelB: I would prefer to use  `variant` for that prop name  with default on "primary"
also why not use flavors ?

UnbearableBear: Agree with variant. With the default on primary however, this would cause regressions. The absence of "variant / kind" is also a style in itself (cf. story book)

UnbearableBear: We could change that in another PR

lionelB: ok, can we also have a variant="clear" or something like which is like the default ?

UnbearableBear: Hum we could have something like `variant="default"` ?

lionelB (approved):

PR #796, opened 2019-05-02
feat(ui): add kind icon

UnbearableBear:
#783 step 2


packages/code-du-travail-ui/src/ToggleButton/doc.md
```diff
@@ -72,6 +74,16 @@ import ToggleButton from ".";
     {" "}
     <ToggleButton kind="danger" disabled pressed>Button Danger</ToggleButton>
   </p>
+  <p>
+    <ToggleButton kind="icon">
```
lionelB, line 78: From an APi point of view, I would export `<ToggleButton kind="Icon />` as ans `<IconButton />` 
what do you think  ?

PR #797, opened 2019-05-02
feat(ui): add toast component

UnbearableBear:
#783 step 3

### DIFF
--- a/packages/code-du-travail-frontend/src/layout/Footer.js
+++ b/packages/code-du-travail-frontend/src/layout/Footer.js
@@ -20,7 +20,7 @@ const Footer = () => (
           vous conseiller et vous orienter dans vos démarches.
         </p>
         <ServiceRenseignementModal>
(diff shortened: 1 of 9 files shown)


packages/code-du-travail-frontend/src/layout/Footer.js
```diff
@@ -20,7 +20,7 @@ const Footer = () => (
           vous conseiller et vous orienter dans vos démarches.
         </p>
         <ServiceRenseignementModal>
-          <ToggleButton primary>
+          <ToggleButton kind="primary">
```
lionelB, line 23: this should be in the ToggleButton PR ?

UnbearableBear: You are absolutely right !

lionelB: THANKS!

PR #800, opened 2019-05-03
feat(ui): and icon variant

UnbearableBear:
I prefer not to have a specific component to handle a simple icon case where we only have to remove a few styles and keep all the rest


lionelB (commented):
do you forget to add tests ?

UnbearableBear:
Fixed !

lionelB (approved):

PR #801, opened 2019-05-03
feat(ui): add new toast component

UnbearableBear:


packages/code-du-travail-ui/src/Toast/index.js
```diff
@@ -0,0 +1,186 @@
...
+      </StyledToast>
+    );
+  }
+}
+
+Toast.propTypes = {
+  /**
+   * Type of toast to show
+   */
+  type: PropTypes.oneOf(["warning", "success", "info"]),
```
lionelB, line 51: I wonder if we should use flavors here ?

packages/code-du-travail-ui/src/Toast/index.js
```diff
@@ -0,0 +1,186 @@
...
+  min-height: 48px;
+  width: ${props => (props.wide ? "100%" : "auto")};
+  background-color: white;
+  border-color: var(--infoColor);
+  border-style: solid;
+  border-width: 1px;
+  border-radius: ${box.borderRadius};
+  ${props => {
+    let animation = "none";
+    let borderColor = colors.infoBackground;
```
lionelB: ` borderColor = colors[`${props.variant}Background`];`

packages/code-du-travail-ui/src/Toast/index.js
```diff
@@ -0,0 +1,186 @@
...
+export default Toast;
+
+const StyledToast = styled.div`
+  box-sizing: border-box;
+  display: inline-flex;
+  justify-content: space-between;
+  min-height: 48px;
+  width: ${props => (props.wide ? "100%" : "auto")};
+  background-color: white;
+  border-color: var(--infoColor);
```
lionelB: not needed

packages/code-du-travail-ui/src/Toast/index.js
```diff
@@ -0,0 +1,186 @@
...
+      borderColor = colors.successBackground;
+    }
+    if (props.animate) {
+      if (props.animate === "from-top") {
+        animation = css`
+          ${fromTop} 0.3s ease-out
+        `;
+      } else if (props.animate === "from-right") {
+        animation = css`
+          ${fromRight} 0.5s ease-out
```
lionelB, line 92: is it normal that top / bottom have not the same duration than left / right ?

UnbearableBear: Yep because we have more distance on left / right animations so we want them to be slightly longer.

packages/code-du-travail-ui/src/Toast/index.js
UnbearableBear: My secret goal is to shoot the flavor utils 🤓

packages/code-du-travail-ui/src/Toast/index.js
UnbearableBear: nice catch !

packages/code-du-travail-ui/src/Toast/index.js
```diff
@@ -0,0 +1,160 @@
...
+  type: PropTypes.oneOf(["warning", "success", "info"]),
+  wide: PropTypes.bool,
+  shadow: PropTypes.bool,
+  animate: PropTypes.oneOf([
+    "from-top",
+    "from-right",
+    "from-bottom",
+    "from-left"
+  ]),
+  timeout: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
```
lionelB: why `PropTypes.oneOfType([PropTypes.bool, PropTypes.number])` and not just `PropTypes.number`

UnbearableBear: Humm seems indeed dumb. I set it to `false` by default. I only have this in the code `    if (timeout) {`  so I guess that if I leave it `undefined` by default instead of `false` it would also do the trick ?

lionelB: yes

lionelB (approved):

PR #809, opened 2019-05-06
feat(frontend): fetch themes related documents

lionelB:
replace the search based on the theme's label with an api call
the theme api retrieve a theme object containing it's documents


UnbearableBear (approved):
LGTM

PR #813, opened 2019-05-07
Tabs component

UnbearableBear:
Tadaaaam ! 
Screenshot 2019-05-07 at 17 46 38

Docz c'est la loose ce truc, j'ai du perdre deux jours à cause d'eux depuis que je suis ici


packages/code-du-travail-ui/src/Tabs/doc.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+import "@cdt/css";
+import { Playground, Props } from "docz";
+import Tabs from ".";
+
+# Tabs
+
+<Playground>
+  <Tabs data={[
```
lionelB: Can we have a children based content rather that do it programmatically ?

packages/code-du-travail-ui/src/Tabs/index.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+Tabs.propTypes = {
+  withHistory: PropTypes.bool,
+  onSelect: PropTypes.func,
+  selectedIndex: PropTypes.number,
+  defaultIndex: PropTypes.number,
+  data: PropTypes.arrayOf(
+    PropTypes.shape({
+      key: PropTypes.string.isRequired,
```
lionelB, line 38: Why key is mandatory ? We could infer it from tab name or tab position

packages/code-du-travail-ui/src/Tabs/index.js
```diff
@@ -0,0 +1,92 @@
...
+      tab: PropTypes.node.isRequired,
+      panel: PropTypes.node.isRequired
+    })
+  ).isRequired
+};
+
+Tabs.defaultProps = {
+  withHistory: false,
```
lionelB: this is interesting :) does it works ? why default it to false ?

packages/code-du-travail-ui/src/Tabs/test.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { render } from "react-testing-library";
+import Accordion from ".";
+
+const items = [{ title: "This is the title", body: "this is the body" }];
+
+describe("<Accordion />", () => {
```
lionelB: I guess it's `<Tabs/>`

packages/code-du-travail-ui/src/Tabs/doc.md
UnbearableBear: Well the issue with this approach is that we must then export several components (`<Tabs>`, `<Tab>`, `<TabList>`, `<TabPanel>`) instead of just one, and this implies to know they exists and how to "order" them which is kind of error prone. And you still have to loop over them to build up your structure (so you end up with two loops/maps at least).  So I'm not sure of the benefits

packages/code-du-travail-ui/src/Tabs/index.js
UnbearableBear: I want to let the possibility to provide customs keys. But then, if I don't make the key mandatory, I could have key duplication if I fallback on the current index and that such key has already been provided or will be.

packages/code-du-travail-ui/src/Tabs/index.js
UnbearableBear: Oh I forgot to remove that one ^^ At first I thought I could handle the push state at the component level but in the end I think it is better to let the parent component handle it if it wishes so.

packages/code-du-travail-ui/src/Tabs/test.js
UnbearableBear: yeeep

packages/code-du-travail-ui/src/Tabs/index.js
UnbearableBear: So I prefer to let the parent component handle all the  key and make them mandatory.

packages/code-du-travail-ui/src/Accordion/index.js
```diff
@@ -47,6 +38,16 @@ class Accordion extends React.PureComponent {
   }
 }
 
+Accordion.propTypes = {
```
UnbearableBear, line 41: @lionelB Tu avais fait quelque chose de spécial pour que ça marche ça ? Parce que ça marche plus de nouveau avec docz quand je fais `yarn start` dans l'ui :(

UnbearableBear: (je parle du `static propTypes`)

adipasquale: pourquoi tu touches a l'Accordion dans cette PR @UnbearableBear ?

adipasquale (approved):

UnbearableBear:
Il faut que je regarde un petit truc en début d'aprem pour essayer de régler ce prob et après je merge !

packages/code-du-travail-ui/src/Tabs/doc.md
lionelB: Yup but this approach is way more composable / customizable

packages/code-du-travail-ui/src/Tabs/index.js
lionelB: Sorry to disagree but I'm not a fan a implementation detail that leak over public compenent api . 
Why do we need to let user define custom key ? We can definitly create some internal ref

packages/code-du-travail-ui/src/Tabs/doc.md
lionelB: To me the default version should export all component, and if you want to add another one wich bundle all the component together using a `data` it's ok too. 
In the end, we will need those component to rendre a flat view

UnbearableBear: But do we want the tabs to be customizable in the frontend ? The role of our UI is to "lock" the styles. About the composability, I'm not sure about what should be composable here ?

packages/code-du-travail-ui/src/Tabs/index.js
UnbearableBear: I agree, let's remove the key then

packages/code-du-travail-ui/src/Tabs/doc.md
lionelB: In order to achieve a flat version of Tabnavigtor 
```
<Tab1>––
–––––––––
<Tab2>––
–––––––––
<Tab3>––
––––––––– 
```
we'll need those components to have the same rendering

packages/code-du-travail-ui/src/Accordion/index.js
lionelB: @UnbearableBear j'ai rien fait de spécial

UnbearableBear: Je vais regarder en début d'aprem: https://github.com/pedronauck/docz/issues/114

lionelB (changes requested):
In addition to the current Tabs components we should export or re-export `Tab`, `TabList`, `TabPanel`

packages/code-du-travail-ui/src/Tabs/doc.md
UnbearableBear: I'm not sure I get it. The package itself won't work this way. So it would imply we make it work and so hacky code for no added value ? In which case we would need that ?

lionelB: The case descibred here was to have a 2 phase render in order to have a complete html page on initial load  (we already discussed that point) that's why we need also need basics components that compose complex components
 
So I'm probably not very clear. I don't mean to add any hacky code,I just  want to reuse the code you use internally in Tabs (styled Tabs / TabPanel) and re-exported them so we can use it I also propose to re-export Tablist so user don't have to bother to import some component from different libs.

UnbearableBear: I'm sorry I really don't get what you want to do. Do you have some time to "proof of concept" it ?

lionelB (approved):

lionelB:
J'aimerai bien qu'on évite à l'avenir de rajouter pleins de truc hors sujet après avoir review une PR.

UnbearableBear:
If you are talking about `refactor(ui): revert md files backs to mdx` It was a naïve attempt to fix the babel issue. But I ended up realizing that keeping the `.mdx` as they were originally is probably better than hacking them to be `.md`
I had not the intention to sneak anything in, nor to take any risk or fool anyone. Just to be sure they are no misunderstanding. It's almost a no-op to me

PR #829, opened 2019-05-15
fix(search): fix IE11 enter key

revolunet:
For some reason the Enter key is inactive on IE11, this force submission on enter key


UnbearableBear (approved):
IE, le running gag

lionelB (commented):
I've updated the PR with a simpler (I hope) solution

revolunet:
better, thanks !

PR #839, opened 2019-05-20
ci(travis): update to Xenial

douglasduteil:
---

Following https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
-dist: trusty
-sudo: false
-language: generic
+dist: xenial
+language: minimal
 git:
   depth: 5
 


lionelB:
can you elaborate a bit on these changes (why)

douglasduteil:
\from https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
> Xenial as the default Travis CI build environment is coming up next week, starting incrementally on Tuesday April 23rd, 2019.

.travis.yml
```diff
@@ -1,6 +1,5 @@
-dist: trusty
-sudo: false
-language: generic
+dist: xenial
```
lionelB, line 1: I wonder if we could just  remove the `dist`, since we always use the default value ?

douglasduteil: :thinking: Let me try

lionelB (approved):

douglasduteil:
Naaaa it's still "trusty" by default :(
see 
https://travis-ci.com/SocialGouv/code-du-travail-numerique/jobs/201812728#L9
vs https://travis-ci.com/SocialGouv/code-du-travail-numerique/builds/112481157#L9

PR #843, opened 2019-05-20
fix(css): minor-issue

UnbearableBear:
the variable was deleted in another PR but this one got below the radar


lionelB (approved):
nice PR split 👍

PR #845, opened 2019-05-21
fix: upgrade sentry setup

revolunet:
update sentry setup with new reference implementation : https://github.com/zeit/next.js/tree/canary/examples/with-sentry-simple

### DIFF
--- a/packages/code-du-travail-frontend/pages/_error.js
+++ b/packages/code-du-travail-frontend/pages/_error.js
@@ -13,17 +13,14 @@ if (typeof window !== "undefined") {
   Sentry.init({ dsn: SENTRY_PUBLIC_DSN, debug: true });
 }
 
-const BrowserException = args => new Error(args);
-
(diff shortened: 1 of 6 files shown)


lionelB (approved):
Thanks!!!

packages/code-du-travail-frontend/pages/_error.js
```diff
@@ -13,17 +13,14 @@ if (typeof window !== "undefined") {
...
 
-const BrowserException = args => new Error(args);
-
-const notifySentry = err => {
+const notifySentry = statusCode => {
   if (typeof window === "undefined") {
     return;
   }
-  Sentry.configureScope(scope => {
+  Sentry.withScope(scope => {
```
lionelB, line 20: we should verify that we don't have duplicated error since I guess _error will be called after an error occur

lionelB (approved):

PR #847, opened 2019-05-21
Convention sidebar auto-tracking

adipasquale:
This PR uses [tocbot](https://github.com/tscanlin/tocbot) to allow automatic tracking of the matching title in the sidebar table of contents, for the currently visible content.

cf https://streamable.com/5djwv for a screencast

features: 
- supports infinite nested levels
- works together with expanding sections. if you explicitly expand a section it will stay expanded.
- responsiveness is handled by disabling the tracking feature when the ToC is displayed on top of the content instead of as a sidebar

does not feature:
- automatic scrolling of the sidebar for long contents

This PR also includes minor styling updates from #912

## TocBot DOM 

here are two HAML examples to show you what kind of DOM is expected by tocbot:

the content: 

```HAML
.js-toc-content
  h3 
    Congés Payés
  h3
    Prévoyance
  h4 
    Adhésion
  h4
    Souscription
```

the table of contents (aka the sidebar):

```HAML
.js-toc
  .toc-list
    .toc-list-item
      .toc-link 
        Congés Payés
    .toc-list-item
      .toc-link
        Prévoyance
      .toc-list
        .toc-list-item
          .toc-link
            Adhésion
        .toc-list-item
          .toc-link
            Souscription
```

We are building the ToC DOM ourselves using React Components that are regrouped
in a `toc` folder.

## Refactor 

If you have a suggestion to refactor this bit to be DRYer I'd be glad :

https://github.com/SocialGouv/code-du-travail-numerique/pull/847/files#diff-37f8a04da8b2c1ac553d8624e93e1a72R22


packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
```diff
@@ -1,22 +1,24 @@
 import React from "react";
 import PropTypes from "prop-types";
-import styled, { css } from "styled-components";
+import styled from "styled-components";
 import { theme } from "@cdt/ui/";
+import { TocListItem, TocList, TocLink } from "./toc";
 
-const SummaryItem = ({ data, children, type, level, visible }) => {
```
adipasquale, line 6: the visible property was actually not used, it was a leftover from a previous version that shouldn't have made it to master

revolunet:
modif mergée upstream 👌https://github.com/tscanlin/tocbot/pull/140

packages/code-du-travail-frontend/src/conventions/texte/toc/index.js
```diff
@@ -0,0 +1,5 @@
+import TocLink from "./TocLink";
+import TocListItem from "./TocListItem";
+import TocList from "./TocList";
+
+export { TocLink, TocList, TocListItem };
```
lionelB: we try avoid that kind of index files and import directly files

packages/code-du-travail-frontend/src/conventions/texte/toc/TocListItem.js
```diff
@@ -0,0 +1,39 @@
...
+    {children}
+  </Wrapper>
+);
+
+const Wrapper = styled.li`
+  list-style-type: none;
+  margin-left: 5px;
+  padding-left: 10px;
+  ${props =>
+    props.level != undefined &&
```
lionelB: you can remove this test by adding a default value to props

packages/code-du-travail-frontend/src/conventions/texte/toc/TocListItem.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import styled, { css } from "styled-components";
+import PropTypes from "prop-types";
+
+const TocListItem = ({ level, type, id, children }) => (
```
lionelB: there are missing PropType / defautProps

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -26,6 +27,38 @@ class ConventionTexte extends React.Component {
...
+    this.resize();
+    window.addEventListener("resize", () => this.resize());
+  }
+
+  componentDidUpdate() {
+    this.mountOrUnmountTocbot();
+  }
+
+  componentWillUnmount() {
+    tocbot.destroy();
```
lionelB, line 42: you should remove `resize` event listener too

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -26,6 +27,38 @@ class ConventionTexte extends React.Component {
...
+  }
+
+  resize() {
```
lionelB: you should probably debounce this call to avoid to many calls

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -26,6 +27,38 @@ class ConventionTexte extends React.Component {
...
+    this.setState({
+      tocbotEnabled: window.innerWidth > parseInt(theme.breakpoints.tablet, 10)
+    });
+  }
+
+  mountOrUnmountTocbot() {
+    const { tocbotMounted, tocbotEnabled } = this.state;
+    if (!!tocbotEnabled == !!tocbotMounted) return;
```
lionelB: this is really weird , can you ensure that these state variables are always boolean ?

packages/code-du-travail-frontend/src/conventions/texte/ContentTitle.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+  } else if (level == 0) {
+    return <TopSectionTitle id={id}>{titre}</TopSectionTitle>;
+  } else {
+    return <NestedSectionTitle id={id}>{titre}</NestedSectionTitle>;
+  }
+};
+
+// TODO: this method should be used in the styled component definitions below
+const getTitleTagName = ({ type, level }) => {
```
lionelB, line 19: in addition to that you can use the `as` on styledComponent
`<MyComponent as="h1" />  `
`<MyComponent as="h2" /> `
`<MyComponent as="h3" /> `

if it can help

packages/code-du-travail-frontend/src/conventions/texte/ContentTitle.js
```diff
@@ -0,0 +1,45 @@
...
+    return "h5";
+  } else if (level == 0) {
+    return "h3";
+  } else {
+    return "h4";
+  }
+};
+
+const TopSectionTitle = styled.h3`
+  font-size: 22px;
```
lionelB: for that values, It would be better to rely on the UI theme values

packages/code-du-travail-frontend/src/conventions/texte/Sidebar.js
```diff
@@ -2,20 +2,24 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Section, Wrapper } from "@cdt/ui";
 import SummaryTitle from "./SummaryTitle";
+import { TocList } from "./toc";
 
-const Sidebar = ({ node, onSummaryTitleToggleExpanded }) => {
+const Sidebar = ({ node, onSummaryTitleToggleExpanded, tocbotEnabled }) => {
   return (
-    <Section>
+    <Section className="js-toc">
```
lionelB, line 9: I guess `js-toc` is  the default value here when initializing Toc, 
Can you add a comment here to let "futur us" know that it tocbot related ?

packages/code-du-travail-frontend/src/conventions/texte/SummaryTitle.js
```diff
@@ -1,37 +1,48 @@
...
-            {child.children}
-          </SummaryItem>
-        ))}
-    </Wrapper>
+    <TocListItem type={type} id={id} level={0}>
+      <TitleLink
+        id={id}
+        type={type}
+        onClick={e => {
+          e.preventDefault();
```
lionelB: to avoid passing a new function to TitleLink to each render, 
you can use a useCallback hook here
```
 const clickHandler = useCallback(event => {
  event.preventDefault();
  onToggleExpanded(id, !expanded);
}, [data.id, expanded])
```

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
```diff
@@ -1,22 +1,31 @@
...
+      {children && (
+        <TocList expanded={expanded}>
+          {children.map(child => (
+            <SummaryItem
+              expanded={expanded}
+              key={child.data.id}
+              level={level + 1}
+              {...child}
+            >
+              {child.children}
```
lionelB: I didn't realize before but I'm very confused by the `children`  props here
React Element have a prop called children too which contains children Reacts Element of a givent ReactElements.

lionelB (changes requested):

packages/code-du-travail-frontend/src/conventions/texte/toc/TocListItem.js
adipasquale: what do you mean ? all 4 params are declared in the propTypes. I added a default value for level as suggested in the previous comment

lionelB: oups....  my bad I didn't see the proptypes

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
adipasquale: very good point thank you 🙇

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
adipasquale: indeed after re-reading this feels weird. I think it's because they are sometimes null. I'll add default values for them to avoid this.

packages/code-du-travail-frontend/src/conventions/texte/ContentTitle.js
adipasquale: that helped a lot I didn't know about the `as` prop ! I refactored this component: https://github.com/SocialGouv/code-du-travail-numerique/commit/b2a5160da07515f349a87e69084fe4c522fb95f6

lionelB: great !

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
adipasquale: yes these are the same `children` I am referencing here, the React props. My understanding is that if you call a component with something inside the tag like :

```jsx
<Title>
  <div>bonjour</div> 
  <span>coucou</div>
</Title>
```
in the Title component, the `children` prop will contain these two subcomponents, the div and the span.

in the SummaryItem, I'm calling the component recursively, so I'm doing something like :

```jsx
const SummaryItem = ({ children }) => (
  <React.Fragment>
  {children.map(child => (
    <SummaryItem>{child.children}</SummaryItem>
  ))}
  </React.Fragment>
);
```

is it weird ? I could also pass down the children with `<SummaryItem children={child.children} />` but I don't feel like it's any better

packages/code-du-travail-frontend/src/conventions/texte/SummaryTitle.js
adipasquale: thanks I didn't know about useCallback, I refactored using it https://github.com/SocialGouv/code-du-travail-numerique/commit/98fe141df2102405c3924d457582f02b75d3941f

adipasquale:
ping pour la re-review @lionelB

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
lionelB: if you pass an anonymous  function to the resize event listener il will not be possible to remove it later 
I would recommand to do 
```js
window.addEventListener("resize", this.resize);
```
and later 
```js
window.removeEventListener("resize", this.resize);
```

adipasquale: yes this is what I did in the refactor after your comments

packages/code-du-travail-frontend/src/conventions/ConventionTexte.js
```diff
@@ -5,16 +5,19 @@ import Sidebar from "./texte/Sidebar";
...
     super(props);
+    const baseState = { tocbotEnabled: false, tocbotMounted: false };
     if (props.preloadedTexte) {
       const texte = props.preloadedTexte;
       const topNode = this.getFirstNodeWithChildren(texte);
-      this.state = { topNode, texte, loaded: true };
+      this.state = Object.assign(baseState, { topNode, texte, loaded: true });
     } else {
-      this.state = { loaded: false };
+      this.state = Object.assign(baseState, { loaded: false });
```
revolunet: pour un peu plus de consistance avec le reste du code on pourrait faire ?

```js
this.state = {
  ...baseState,
  loaded: false
}
```

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
revolunet: btw i think you should use `React.Children.map` https://reactjs.org/docs/react-api.html#reactchildren

adipasquale: nice I didn't know about that I'll try @revolunet

adipasquale: actually I think that the main problem was that I was mixing React's `children` and my own AST-like tree structure's `children`

adipasquale: https://github.com/SocialGouv/code-du-travail-numerique/pull/847/commits/c5a33f8e99e45d7fb75da0586c59be0d90dd3ef9 @revolunet I think it's much clearer (even for me 😬 )

packages/code-du-travail-frontend/src/conventions/texte/SummaryTitle.js
```diff
@@ -23,7 +17,7 @@ const SummaryTitle = ({
     [id, expanded]
   );
 
-  if (!children || children.length == 0) {
+  if (!node.children || node.children.length == 0) {
```
lionelB: it should be '==='

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
       dockerfile: ./Dockerfile
 
   nlp_api:
-    restart: always
+    # restart: always
```
lionelB: is this change ok ?

packages/code-du-travail-frontend/src/conventions/texte/SummaryItem.js
lionelB: thanks @adipasquale for the renaming 🙏

docker-compose.yml
adipasquale: woups I didn't mean to commit this 🙄 I'm doing this locally because I don't know how to set it up and so it keeps restarting

lionelB (approved):

PR #850, opened 2019-05-22
fix(frontend): issue in matomo

### LINKED ISSUE #722: matomo: dynamic action_name
Matomo utilise le titre de la 1ère page chargée en tant que `action_name` pour les évènements ultérieurs. ceci doit être dû au mode SSR/[SPA](https://fr.wikipedia.org/wiki/Application_web_monopage)

Voir si on doit utiliser [l'API JS](https://developer.matomo.org/api-reference/tracking-javascript) pour fixer ça, ex `setDocumentTitle`

UnbearableBear:
closes #722

Little explanation: 

`ReactPiwik.push(["trackPageView"]);` must be called only once after the instantiation process.
We could then use the method `connectToHistory` provided by the instance of the piwik class but it seems to require an instance of `react-router` (thanks for pointing that out @lionelB). So instead, I use the listener of our router to call the `track` method of the piwik instance (which is normally called on route change once `connectToHistory` has been called). The `track` method expects a loc object which, in fact, only needs one key, which is: `path`.

https://github.com/guillaumeparis2000/react-piwik/blob/master/src/React-Piwik.js
https://developer.matomo.org/api-reference/tracking-javascript

### DIFF
--- a/packages/code-du-travail-frontend/src/piwik.js
+++ b/packages/code-du-travail-frontend/src/piwik.js
@@ -2,21 +2,20 @@ import ReactPiwik from "react-piwik";
 import { Router } from "../routes";
 import getConfig from "next/config";
 
-const onRouteChangeComplete = () => {
-  ReactPiwik.push(["trackPageView"]);
-};
-
 const {
   publicRuntimeConfig: { PIWIK_URL, PIWIK_SITE_ID }
 } = getConfig();
 
 if (typeof window !== "undefined" && PIWIK_URL && PIWIK_SITE_ID) {
-  new ReactPiwik({
+  const piwik = new ReactPiwik({
     url: PIWIK_URL,
     siteId: PIWIK_SITE_ID,
     trackErrors: true
   });
-  Router.events.on("routeChangeComplete", onRouteChangeComplete);
+  ReactPiwik.push(["trackPageView"]);
+  Router.events.on("routeChangeComplete", path => {
+    piwik.track({ path });
+  });
 } else if (typeof window !== "undefined") {
   window._paq = [];
 }


packages/code-du-travail-frontend/src/piwik.js
```diff
@@ -2,20 +2,26 @@ import ReactPiwik from "react-piwik";
 import { Router } from "../routes";
 import getConfig from "next/config";
 
-const onRouteChangeComplete = () => {
-  ReactPiwik.push(["trackPageView"]);
+let piwik;
```
UnbearableBear: I don't like this piwik here. I wouldn't need it if the callback is directly declared where it's used ?

packages/code-du-travail-frontend/src/piwik.js
```diff
@@ -2,21 +2,20 @@ import ReactPiwik from "react-piwik";
...
 
 if (typeof window !== "undefined" && PIWIK_URL && PIWIK_SITE_ID) {
-  new ReactPiwik({
+  const piwik = new ReactPiwik({
     url: PIWIK_URL,
     siteId: PIWIK_SITE_ID,
     trackErrors: true
   });
-  Router.events.on("routeChangeComplete", onRouteChangeComplete);
+  ReactPiwik.push(["trackPageView"]);
```
lionelB, line 15: why not call `piwik.track({path: window.location.pathname})`

packages/code-du-travail-frontend/src/piwik.js
```diff
@@ -2,21 +2,20 @@ import ReactPiwik from "react-piwik";
...
+  Router.events.on("routeChangeComplete", path => {
+    piwik.track({ path });
```
lionelB, line 17: path should not be a complete url `http://{hostname}/{pathname}` but a pathname since 
the Piwik.track will append `${window.location.origin}`
What do you think of using track(window.location) wich will fallback to (pathname + search)

UnbearableBear: It turns out the path returned by the router is exactly pathname + search !

lionelB: the nextjs documentation is a bit misleading :(
> routeChangeComplete(url) - Fires when a route changed completely
> Here url is the URL shown in the browser. If you call Router.push(url, as) (or similar), then the value of url will be as.

https://github.com/zeit/next.js#router-events

UnbearableBear: Plus we also have a package around the router, it might changes a few things too. I didn't read the doc, I simply `console.log` it to see the value.

lionelB (approved):

PR #852, opened 2019-05-22
refactor(frontend): add hostname to feedbackform

### LINKED ISSUE #840: Feedbacks en environnement de test
Aujourd'hui les formulaires du module de feedback sont envoyés sur tous les environnements ce qui fausse les stats quand on fait des tests. On peut envisager 2 possibilités :
- soit on envoie un paramètre supplémentaire qui permet de distinguer les environnements et on filtre dans l'étude
- soit on envoie les formulaires que lorsqu'on est sur la version de prod

UnbearableBear:
closes #840

### DIFF
--- a/packages/code-du-travail-frontend/src/common/Feedback/feedback.service.js
+++ b/packages/code-du-travail-frontend/src/common/Feedback/feedback.service.js
@@ -8,7 +8,7 @@ export function postFeedback(url, data) {
       Accept: "application/json",
       "Content-Type": "application/json"
     },
-    body: JSON.stringify({ ...data, hostname: window.location.hostname })
+    body: JSON.stringify(data)
   })
     .then(r => r.json())
     .then(data => {
--- a/packages/code-du-travail-frontend/src/common/Feedback/feedback.service.js
+++ b/packages/code-du-travail-frontend/src/common/Feedback/feedback.service.js
@@ -8,7 +8,7 @@ export function postFeedback(url, data) {
       Accept: "application/json",
       "Content-Type": "application/json"
     },
-    body: JSON.stringify(data)
+    body: JSON.stringify({ ...data, hostname: window.location.hostname })
   })
     .then(r => r.json())
     .then(data => {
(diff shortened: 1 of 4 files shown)


revolunet:
Rename le message de commit : pathname -> hostname ?

packages/code-du-travail-frontend/src/common/Feedback/feedback.service.js
```diff
@@ -8,7 +8,7 @@ export function postFeedback(url, data) {
       Accept: "application/json",
       "Content-Type": "application/json"
     },
-    body: JSON.stringify(data)
+    body: JSON.stringify({ ...data, hostname: window.location.hostname })
```
lionelB: I would prefer to add  `hostname` where we gather the data

lionelB: also, I thought we already have the host in data to send...

UnbearableBear: Well now that you make me think about it, we do send an `url`param which is `pathname + query` and we put in every config data object. Maybe we should send the complete path in one param here and that's it.

lionelB: maybe we should ask @ClementChapalain if he prefers to have it splited or not... 
Yet, I think hostname should be added to data at the same place other data are added too

UnbearableBear: IMHO, the issue is a global one here, whatever the call. We want all the forms to have the env so we can check wether they are from localhost dev etc. So it would be error prone to manually add it in data everytime

ClementChapalain: as long a I can easily split the env afterwards in the spreadsheet (with a simple regex for example : https://support.google.com/docs/answer/3098244?hl=fr) it's ok for me :)

lionelB: So I would say go send the compte url ?

UnbearableBear: Ok so I added the complete URL every time we call the postFeedback. And I made sure it is only used when we know it's not in an SSR context (i.e. in user action's callback only)

lionelB (approved):

PR #864, opened 2019-05-28
refactor(chore): add root linter

UnbearableBear:
C'est une bonne idée ça ?


lionelB (commented):
there is obviously some changes that aren't related to moving a file

UnbearableBear:
These are files that got formated because of the linting changes. But it seems that the CI is not liking that

lionelB:
I think it will be more usefull to add a generic config a the top and ovverife it in project (ie a11y rules are only usefull in frontend)

PR #865, opened 2019-05-29
refactor(chore): add root linter

UnbearableBear:


.eslintrc.yml
```diff
@@ -0,0 +1,7 @@
+root: true
+extends:
+  - "@socialgouv/eslint-config-recommended"
+
+rules:
+  prefer-const: error
+  no-console: off
```
lionelB: we don't want console (we only tolerate it on frontend but we should move them to sentry log)

UnbearableBear: Hum the issue is that most of our files in the dataset folder use `console.log` to pipe the output into files with `>>`

lionelB: what if we add "root" eslint config in `code-du-travail-data` where we override the use of console ?

UnbearableBear: Makes total sense !

UnbearableBear: done !

lionelB (approved):

PR #866, opened 2019-05-29
feature(data): add fiche sp fetcher

UnbearableBear:
Any thoughts on that ?


packages/code-du-travail-data/dataset/fiches_service_public/format.js
```diff
@@ -1,4 +1,4 @@
-const parseReference = require("./parseReference");
+const parseReference = require("./parse-reference");
```
lionelB: why did you rename file here ? most of the filename are camelCase

lionelB (commented):
Don't see why you revert the filenames to camelCase since most of script are camelCase here

lionelB:
I wonder if using bash would be a better in that it's not tied to node env ? 
but  👍  !

UnbearableBear:
Maybe in a future version (i'm not very familiar with bash processes) but all the parsing is based on node right now so I guess it's ok

packages/code-du-travail-data/dataset/fiches_service_public/format.js
UnbearableBear: Consistency, some files were camelCased some others not so I picked the most common (in this folder). I always prefer files not to be camel cased though. Old habitude, I find it slighty more readable, and by convention "-" means "these are two distinct words".

lionelB:
@UnbearableBear est ce que tu dé-renomer les fichiers ? le kebab-case c'est utilisé pour les fichier json les scripts sont en camelCase

UnbearableBear:
@lionelB je m'en occupe

lionelB (approved):

PR #872, opened 2019-06-03
refactor(ui): remove css of tag component

UnbearableBear:
Un avis avant que je n'applique le reste à l'alert et au badge ?


packages/code-du-travail-ui/src/Button/index.js
```diff
@@ -6,8 +6,12 @@ import { darken, lighten, transparentize } from "polished";
 
 // We don't want the variant prop to be passed down to the button
 // eslint-disable-next-line
-const Button = ({ pressed, variant, onClick, ...props }) => {
-  return <button aria-pressed={pressed} {...props} onClick={onClick} />;
+const Button = ({ children, pressed, variant, onClick, ...props }) => {
```
lionelB, line 9: Is there a difference between spreading props with / without children ?

UnbearableBear: I don't think so, except it is slighlty more explicit ? If we one day decide to remove the `{...props}` children will still be there.

lionelB (approved):

PR #895, opened 2019-06-05
feat: add page /stats for #873

revolunet:
En attente des chiffres :)

### DIFF
--- a/packages/code-du-travail-frontend/pages/stats.js
+++ b/packages/code-du-travail-frontend/pages/stats.js
@@ -18,7 +18,6 @@ const About = () => (
         <Wrapper variant="light">
           <h1>Statistiques du code du travail numérique</h1>
           <br />
-          <br />
           <Grid>
             <GridCell>
               <h3>Contenus référencés</h3>
@@ -38,7 +37,7 @@ const About = () => (
             </GridCell>
           </Grid>
           <br />
-          <br />
+          <p>Statistiques d’usage depuis le 1er Janvier 2019</p>
           <br />
           <br />
         </Wrapper>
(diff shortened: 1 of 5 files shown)


revolunet:
thanks @UnbearableBear :)

revolunet:
J'ai mis les 4 chiffres réels que j'ai pu trouver (extrait depuis piwik depuis 01/01/2019)

lionelB:
@revolunet il faudrait rajouter les unités et peut être une date si c'est cumulé.

lionelB (approved):
:shipit:

packages/code-du-travail-frontend/pages/stats.js
```diff
@@ -0,0 +1,53 @@
...
+    </Head>
+    <Section>
+      <Container>
+        <Wrapper variant="light">
+          <h1>Statistiques du code du travail numérique</h1>
+          <br />
+          <Grid>
+            <GridCell>
+              <h3>Contenus référencés</h3>
+              <Num>14369</Num>
```
UnbearableBear, line 24: Is it because of the limitation in the matomo api that we don't fetch the data dynamically on each page load here ? Could we store them in a file (as database) to avoid having to update them by hand ? I'd be glad to do it !

lionelB: Ideally, we could fetch it from API which fetch it / cache it from matomo and build graph upon it. I feel we should focus on features with better impact

UnbearableBear: Yeah no big deal !

PR #927, opened 2019-06-12
refactor(css): migrate global styles to frontend app

UnbearableBear:
We are almost done with this css subrepo, almost !


lionelB (changes requested):
@UnbearableBear can you add snapshots ?

UnbearableBear:
@lionelB You mean screenshots ? because they are no diff in the snapshots

lionelB:
@UnbearableBear you ned to add a snapshot for the GlobalStyle component

UnbearableBear:
@lionelB Nice catch, I added it (even if it's empty) and I also split the global style in smaller files. I had an issue with styled components and the inline svg so I used a base64 encoding to avoid that https://github.com/styled-components/jest-styled-components/issues/97

lionelB (approved):

PR #941, opened 2019-06-14
feat(frontend): CCNs display UI improvements

adipasquale:
related to #916


packages/code-du-travail-frontend/src/conventions/ConventionListeTextes.js
```diff
@@ -64,8 +64,9 @@ class ConventionListeTextes extends React.Component {
 
 const Button = styled.button`
   padding: 0;
-  margin: 0;
+  margin: 0 0 8px 0;
```
lionelB: can you use a value from spacing's theme ?

packages/code-du-travail-frontend/pages/kali.js
```diff
@@ -50,8 +49,8 @@ class Kali extends React.Component {
           title={title}
           emptyMessage="Cette convention collective n'a pas été trouvée"
           footer="Informations fournies par la DILA"
-          sourceType="Convention collective"
-          icon={ArticleIcon}
+          icon={null}
```
lionelB: why not removing the icon prop ?

packages/code-du-travail-frontend/pages/kali.js
```diff
@@ -50,8 +49,8 @@ class Kali extends React.Component {
...
+          narrow={false}
```
lionelB: same here bool value in jsx (and html) 
no prop =>  false 
`narrow` => true

and in Answer component,  narrow="false" to be sure it default to no-value

packages/code-du-travail-frontend/pages/kali.js
adipasquale: it's because there is a default value here https://github.com/SocialGouv/code-du-travail-numerique/blob/b4ff54cbf17f7093881536c8789d91b0cc453afe/packages/code-du-travail-frontend/src/common/Article.js#L12 so I have to explicitly say I don't want this default value

packages/code-du-travail-frontend/pages/kali.js
```diff
@@ -49,8 +49,7 @@ class Kali extends React.Component {
           title={title}
           emptyMessage="Cette convention collective n'a pas été trouvée"
           footer="Informations fournies par la DILA"
-          icon={null}
-          narrow={false}
+          wide={true}
```
lionelB: remarque qui n'amène pas a modification  => les propriétés de type boolean peuvent aussi s'écrire de manière concise `wide`plutot que `wide={true}` (comme en html)

lionelB (approved):
thanks !

PR #965, opened 2019-06-25
WIP Page de liste des conventions collectives

adipasquale:
cf https://mattermost.num.social.gouv.fr/default/pl/a3ef36tntfgebf1dyohbgu8g3y

This is only a PoC, not at all decided if we want to go in that direction

Screenshot 2019-06-25 at 10 04 42


revolunet:
should be fixed by #1080

PR #980, opened 2019-06-28
Step 1 avant d'attaquer la CCN-1486, simplification de l'index / wizard du calculateur

UnbearableBear:
Pour éviter d'avoir index / wizard qui gèrent des states différents mais liés, avec du onChange et des passages de prop, j'ai rajouté l'index de la step dans le step reducer et j'ai supprimé le fichier wizard.js
Plus besoin d'initial values dans le form et de useState un peu partout !

Le plus important se trouve ici https://github.com/SocialGouv/code-du-travail-numerique/pull/980/commits/be59cf226990b7c85e198d5ee1b2616a13fbf82f

Au final le fichier index.js ressemble à ça:
https://github.com/SocialGouv/code-du-travail-numerique/blob/1903c689e3f2f5c2c8c0ece4a75a57537c477078/packages/code-du-travail-frontend/src/outils/indemniteLicenciement/index.js

Les deux autres commit j'ai simplement fait de renaming / déplacement bête et méchant


packages/code-du-travail-frontend/src/outils/indemniteLicenciement/PrevNextBar.js
```diff
@@ -3,7 +3,9 @@ import PropTypes from "prop-types";
 import styled from "styled-components";
 import { Button, theme } from "@cdt/ui";
 
-function PrevNextBar({ onPrev, disabled, nextVisible, previousVisible }) {
+function PrevNextBar({ currentStepIndex, disabled, onPrev, stepsLength }) {
```
lionelB, line 6: I don't see the point of changing that, It implies to use PrveNextBar only in a model of with index base navigation :/

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/StepNav.js
```diff
@@ -5,18 +5,18 @@ import { theme } from "@cdt/ui";
  * For now, only <Step /> are used but stepItems will
  * allows to navigate between form's steps
  */
-export function StepItems({ activeIndex = 0, items = [] }) {
+export function StepNav({ activeIndex, steps }) {
```
lionelB, line 8: For now it's not a step navigation, we don't know if we want to do that. That why it was named StepItems and not StepNavigationItems as it was the case at the start

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/Wizard.js
```diff
@@ -1,102 +0,0 @@
-import React, { useState } from "react";
```
lionelB, line 1: I think it's not a good idea to remove the wizzard since we will have several tools that will be based on it. Otherwise we will duplicate most of the wizzard logic on each tool

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/__tests__/indemnite.test.js
```diff
@@ -190,7 +190,7 @@ const tests = [
       dateNotification: "2017-11-01",
       absencePeriods: [
         {
-          type: "Congé parental d'éducation",
+          type: "Congé parental d’éducation",
```
lionelB, line 193: 👍

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/ccn/0016/CategoriePeriod.js
```diff
@@ -1,5 +1,9 @@
 import React from "react";
-import { SectionTitle, Input, InlineError } from "../../stepStyles";
+import {
+  SectionTitle,
+  Input,
+  InlineError
+} from "../../components/styledComponents";
```
lionelB, line 6: I would prefer to call it uiComponents that styledComponents which is tied to a library name

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/ccn/3043/indemnite.js
```diff
@@ -1,6 +1,6 @@
 import { round } from "../../math";
 /**
- * calcul de l'indemnite de licenciement conventionnelle
+ * calcul de l’indemnite de licenciement conventionnelle
```
lionelB, line 3: this looks like a wild search/replace :) Not sure it's a good idea to use typography quote in my comments  :)

lionelB (changes requested):
I think we should more talk about that kind of refactor before do it effectivelly to ensure we share the same vision. To me it make no sense to remove the wizard as it was desiged to be resused over different tool. It's probably not perfect and could be improve for sure.

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/PrevNextBar.js
UnbearableBear: I prefer to keep the logic located only where it's used. During the reducer refactoring I realised that the functions nextVisible and previousVisible were only used un PrevNextBar so I transfered it there

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/StepNav.js
UnbearableBear: StepItems seemed confusing to me

packages/code-du-travail-frontend/src/outils/indemniteLicenciement/Wizard.js
UnbearableBear: Ok I didn't know we planned to reuse it 😄

UnbearableBear:
I guess that the fact the wizard is planned to be reused defeat the purpose of this proposal, let's close it !

PR #982, opened 2019-07-01
fix(data): improve cdtn synonyms syntax

lionelB:
update synonyms

we need a better looks at it from @virginielastisneres @jrduscher and nadia
I've added pass-navigo :) and we can should add more of this kind

cf #984


packages/code-du-travail-data/dataset/synonyms/index.js
```diff
@@ -8,8 +8,20 @@ const thesaurusSynonyms = thesaurus
...
+  const [abbrev, ...tokens] = str
+    .split(",")
+    .sort((a, b) => a.length > b.length)
+    .map(t => t.trim());
+  return `${[...tokens, abbrev].join(", ")} => ${abbrev}`;
+}
+
 const synonyms = [...cdtn_synonyms, ...thesaurusSynonyms]
-  .map(str => str.toLowerCase())
+  .slice(0, 100)
```
revolunet: slice ?

lionelB: oh-noes

revolunet (approved):

PR #990, opened 2019-07-02
fix(frontend): fix undefined meta

### LINKED ISSUE #981: meta are undefined
**Description**
Description claire et concise du bug

**Reproduction**
Etapes à suivre pour reproduire le bug
1. go Home
2. Click on links
3. Inspect head>meta 
4. og:xxxx meta are undefined

**Comportement attendu**
_optionnel_ Si nécessaire, explications du comportement attendu

**Screenshots**
Capture d’écran 2019-06-28 à 17 57 15

UnbearableBear:
fix #981 

1 - When not on a server-side rendered page, url and image in the head tags are undefined
2 - It seems that useEffect does not update the rendering in the `<Head>` component provided by Next.js.

Fix: 
Remove useEffect which is of no use and does not break tests anymore now that Router has been mocked.

### DIFF
--- a/packages/code-du-travail-frontend/pages/_app.js
+++ b/packages/code-du-travail-frontend/pages/_app.js
@@ -29,16 +29,15 @@ export default class MyApp extends App {
     if (Component.getInitialProps) {
       pageProps = await Component.getInitialProps(ctx);
     }
-    // pageUrl is only defined on serverside request
-    pageProps.pageUrl = ctx.req
-      ? `${ctx.req.protocol}://${ctx.req.headers.host}${ctx.req.path}`
-      : undefined;
-
-    pageProps.ogImage =
-      ctx.req &&
-      `${ctx.req.protocol}://${
+    // pageUrl and ogImage are only defined on serverside request
+    if (ctx.req) {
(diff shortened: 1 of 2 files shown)


lionelB:
I think it's safe to replace it it at both server / client render (like it was before you refactor it) to avoid inconsistency on frontend 
Not a bigdeal for the og:image but could be a problem for og:url

UnbearableBear:
They are three things that prevent me from rolling back to previous refactor:

- The first one is about react in general. If a component can define its default props instead of delegating that to a parent component, then I think it should do it. That's because there could be different parent components and thus, create inconsistencies. So I would say that I disagree with you when you say that defining some defaults (among which an unecessary `undefined`) in a parent component is a good solution.
- The second thing is that now that we have tests on pages, the fact that url and image where `required` was causing a lot of red warning while running tests and polluted the output. Which is not the case with the changes you want me to rollback.
- Finally, the last point is that what caused the `undefined` was solely due to the `useEffect` that I added previously. Which is not riquired anymore. So I did already kind of `rollback` ?

Is there anything else that I am missing or would have misunderstood ? I'm not sure I really understand what motivates you in wishing this rollback ?

lionelB (approved):

PR #999, opened 2019-07-03
 refactor(frontend): use typo apostrophe

UnbearableBear:
It's all in the title !

### DIFF
--- a/packages/code-du-travail-frontend/src/outils/indemniteLicenciement/StepItems.js
+++ b/packages/code-du-travail-frontend/src/outils/indemniteLicenciement/StepItems.js
@@ -3,7 +3,7 @@ import styled from "styled-components";
 import { theme } from "@cdt/ui";
 /**
  * For now, only <Step /> are used but stepItems will
- * allows to navigate between form's steps
+ * allows to navigate between form’s steps
  */
 export function StepItems({ activeIndex = 0, items = [] }) {
   return (
--- a/packages/code-du-travail-frontend/src/outils/indemniteLicenciement/Wizard.js
+++ b/packages/code-du-travail-frontend/src/outils/indemniteLicenciement/Wizard.js
@@ -39,7 +39,7 @@ function Wizard({
     } else {
       /**
        * we update initialValues so we can reset the property once form is finished
-       * if we don't update initialValues, initialValues will still equal to an empty object
+       * if we don’t update initialValues, initialValues will still equal to an empty object
        * and updating to another empty object will not trigger form values udate
        */
       setValues(values);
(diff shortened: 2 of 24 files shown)


packages/code-du-travail-frontend/src/outils/indemniteLicenciement/StepItems.js
```diff
@@ -3,7 +3,7 @@ import styled from "styled-components";
 import { theme } from "@cdt/ui";
 /**
  * For now, only <Step /> are used but stepItems will
- * allows to navigate between form's steps
+ * allows to navigate between form’s steps
```
UnbearableBear, line 6: Is it a problem if we have real apostrophe in comments ?

revolunet: no idea !

lionelB: I  don’t use it since it’s need a combination of keys and I find it not to type compare to apostrophe. Also it’s probably a French typographical specificity since most English comment don’t use typographical quote or apostrophe.

UnbearableBear: The issue is that it's hard to replace them automatically if I must make a distinction between the apostrophe in the text and in comments. But it's not specific to french typography !
(https://en.wikipedia.org/wiki/Apostrophe [typographic form])
```Both simplifications carried over to computer keyboards and the ASCII character set. However, although these are widely used due to their ubiquity and convenience, they are deprecated in contexts where proper typography is important.```

UnbearableBear: But they are not mendatory at all so we can skip this PR I would still manage to sleep 😄

lionelB: >  they are deprecated in contexts where proper typography is important.
JSX => OK
Comment => NOK 

aussi il faudrait plutot voir comment eslint ne pourrait pas le faire à notre place

UnbearableBear: Ah mais complètement ! J'espère qu'il y a une règle pour ça, je vais check

UnbearableBear: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/327 C'est tout ce que j'ai trouvé. C'est vrai que les false positive peuvent être nombreux. Donc bon, on dirait bien que ça n'existe pas pour le moment et que si un jour c'est mergé dans ce proj, ce sera "juste" un warning.

PR #1006, opened 2019-07-05
Lionelb/fix error modeles page

lionelB:
add `withError` Hoc to handle gracefully 4xx-5xx requests.
It allow to swap component rendering depending on the response.ok flag
In case of a fail request, it will display the _errorPage and let sentry
do its job.


packages/code-du-travail-frontend/pages/fiche-service-public.js
```diff
@@ -21,7 +22,7 @@ const {
 } = getConfig();
 
 const fetchFiche = ({ slug }) =>
-  fetch(`${API_URL}/items/fiches_service_public/${slug}`).then(r => r.json());
+  fetch(`${API_URL}/items/fiches_service_public/${slug}`);
```
revolunet: no more json ?

revolunet:
looks great, but shouldn't this happen at `_app.js` level instead of every single page ?

lionelB:
@revolunet it could, in that case I can remove the withErrorHoc and do the branching in app directly

revolunet:
Awesome !

revolunet (approved):

packages/code-du-travail-frontend/pages/_error.js
```diff
@@ -10,30 +10,33 @@ const {
   publicRuntimeConfig: { SENTRY_PUBLIC_DSN }
 } = getConfig();
 
-if (typeof window !== "undefined") {
+if (typeof window !== "undefined" && SENTRY_PUBLIC_DSN) {
   Sentry.init({ dsn: SENTRY_PUBLIC_DSN, debug: true });
```
UnbearableBear, line 14: C'est normal qu'on ait un sentry init deux fois (un dans _app et un dans _error) ?

lionelB: ca permet d'attraper les erreur dans app qui interviennent au moment du  render (componentDidCatch)

PR #1008, opened 2019-07-08
Lionelb/indexing crawler bot

lionelB:
This PR allow to handle crawling bots depending on the ENVIRONMENT CI flag.
- [x] block all content from being indexed by bot on dev / staging deployement
- [x] instruct bot to noindex but follow search results pages using meta
- [x] add a canonical meta on content coming from other sources

### DIFF
--- a/packages/code-du-travail-frontend/pages/fiche-service-public.js
+++ b/packages/code-du-travail-frontend/pages/fiche-service-public.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { withRouter } from "next/router";
+import Head from "next/head";
 import fetch from "isomorphic-unfetch";
 import styled from "styled-components";
 import getConfig from "next/config";
@@ -66,6 +67,9 @@ class Fiche extends React.Component {
           description={description}
           image={ogImage}
         />
+        <Head>
+          <link href={data._source.url} rel="canonical" />
+        </Head>
         <ServicePublic>
           <Answer
             title={title}
--- a/packages/code-du-travail-frontend/server.js
+++ b/packages/code-du-travail-frontend/server.js
@@ -3,18 +3,50 @@ const express = require("express");
 const next = require("next");
 const routes = require("./routes");
 
+/**
+ * this env variable is use to target developpement / staging deployement
+ * in order to block indexing bot using a x-robot-header and an appropriate robots.txt
+ */
+const LIVEPROD = process.env.ENVIRONMENT === "production";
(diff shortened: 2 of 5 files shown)


packages/code-du-travail-frontend/server.js
```diff
@@ -3,18 +3,50 @@ const express = require("express");
 const next = require("next");
 const routes = require("./routes");
 
+/**
+ * this env variable is use to target developpement / staging deployement
+ * in order to block indexing bot using a x-robot-header and an appropriate robots.txt
+ */
+const LIVEPROD = process.env.ENVIRONMENT === "production";
```
UnbearableBear, line 10: Don't we use the NODE_ENV variable ? I thought it was some sort of convention

lionelB: NODE_ENV is generally used to build a "production" bundle
we use production bundle for  dev / staging environment
What I need is to know if the env is the production site (aka codedutravail.num.social.gouv.fr) or any other deployement (aka codedutravail-dev.num.social.gouv.fr, xxxxxx.code-du-travail-numerique.dev.factory.social.gouv.fr) to handle the robots.txt accordingly

packages/code-du-travail-frontend/pages/fiche-service-public.js
```diff
@@ -66,6 +67,9 @@ class Fiche extends React.Component {
           description={description}
           image={ogImage}
         />
+        <Head>
+          <link href={data._source.url} rel="canonical" />
```
UnbearableBear, line 71: Since you disallow robots on dev env do we really need the canonical tag ?

UnbearableBear (approved):

packages/code-du-travail-frontend/pages/fiche-service-public.js
lionelB: the canonical stuff is used to instruct search engine that the content here come from another site.

UnbearableBear: So you are telling google that it should index pages on the other domains and not ours ? That's very opinionated in my opinion ! Is that really what we want ?

lionelB: > So you are telling google that it should index pages on the other domains and not ours ? That's very opinionated in my opinion ! Is that really what we want ?

I don't intend to says that. Just that the content is coming from another site

PR #1012, opened 2019-07-09
refactor(data): move synonyms inside the ES image

lionelB:
this PR embed synonyms inside the ES docker image (and we can also do it for stop_words)
to do that, I move the ES Dockerfile inside the data directory so I can later copy dataset files.

Lionel


lionelB:
close since synonyms will be upload to elasticcloud

PR #1019, opened 2019-07-10
fix(frontend): intempestive re-render of form

UnbearableBear:
I created a related issue here for the remainings: https://github.com/SocialGouv/code-du-travail-numerique/issues/1021

### DIFF
--- a/packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
+++ b/packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
@@ -98,14 +98,14 @@ Wizard.propTypes = {
   stepReducer: PropTypes.func.isRequired,
   initialSteps: PropTypes.arrayOf(
     PropTypes.shape({
-      component: PropTypes.element.isRequired,
+      component: PropTypes.func.isRequired,
       name: PropTypes.string.isRequired,
       label: PropTypes.string.isRequired
     })
   ).isRequired,
   initialStepIndex: PropTypes.number,
   initialValues: PropTypes.object,
-  Rules: PropTypes.element
+  Rules: PropTypes.func
 };
 
 export { Wizard };
--- a/packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
+++ b/packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useReducer } from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import { Form } from "react-final-form";
@@ -9,14 +9,14 @@ import { StepItems } from "./StepItems";
 import { PrevNextBar } from "./PrevNextBar";
 
 function Wizard({
-  initialValues = {},
+  initialSteps,
   initialStepIndex = 0,
-  steps,
-  onSubmit,
-  onUpdate,
-  rules = null
+  initialValues = {},
+  Rules = null,
+  stepReducer
 }) {
   const [stepIndex, setStepIndex] = useState(initialStepIndex);
+  const [steps, dispatch] = useReducer(stepReducer, initialSteps);
 
   const prevStep = () => {
     setStepIndex(Math.max(0, stepIndex - 1));
@@ -36,12 +36,13 @@ function Wizard({
   const handlePageSubmit = (values, form) => {
     // This means the user clicked on a "restart a new simulation" button
     if (stepIndex === steps.length - 1) {
-      onSubmit(values);
       form.reset();
+      dispatch({
+        type: "reset"
+      });
       setStepIndex(0);
     } else {
       nextStep();
-      onUpdate && onUpdate(values);
     }
   };
   const stepItems = steps.map(({ name, label }) => ({
@@ -71,7 +72,9 @@ function Wizard({
           return (
             <>
               <form onSubmit={handleSubmit}>
-                {rules}
+                {Rules && (
+                  <Rules values={form.getState().values} dispatch={dispatch} />
+                )}
                 <StepItems activeIndex={stepIndex} items={stepItems} />
                 <StepWrapper narrow noPadding>
                   <Step form={form} />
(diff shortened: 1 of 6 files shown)


packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/index.js
```diff
@@ -1,60 +1,51 @@
...
     <Section>
       <Container>
         <Wrapper size="large" variant="light">
           <h1>Calculateur d&apos;indemnités de licenciement</h1>
           <Wizard
-            steps={steps}
-            onSubmit={onSubmit}
-            onUpdate={setValues}
-            rules={rules}
+            stepReducer={stepReducer}
```
UnbearableBear, line 44: So the main idea of the fix is to delegate the handling of the steps to the wizard. So now, the steps are managed inside of it and not above it, and therefore fix the rerendering issue when changing them.

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/index.js
```diff
@@ -1,60 +1,51 @@
...
-      {async value => {
-        if (value) {
-          const module = await import(`./ccn/${value}`);
-          const steps = module.steps.filter(({ condition = () => true }) =>
-            condition(values)
-          );
-          dispatch({ type: "add_branche", payload: steps });
-        } else {
-          dispatch({ type: "remove_branche" });
+  const Rules = ({ values, dispatch }) => (
```
UnbearableBear, line 13: to make sure we have access to values and dispatch which are in the wizard below, I had to transform `rules` into a component

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/index.js
UnbearableBear: But we can still configure it as before by providing initialStepIndex, InitialValues, initialSteps and rules.

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
```diff
@@ -9,14 +9,14 @@ import { StepItems } from "./StepItems";
...
   initialStepIndex = 0,
-  steps,
-  onSubmit,
-  onUpdate,
-  rules = null
+  initialValues = {},
+  Rules = null,
+  stepReducer
 }) {
   const [stepIndex, setStepIndex] = useState(initialStepIndex);
```
lionelB, line 18: Can we merge this state part into the reducer ?

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/stepReducer.js
```diff
@@ -5,41 +5,39 @@ import { StepAnciennete } from "./steps/Anciennete";
...
+];
 
-export function StepReducer(steps, action) {
+export function stepReducer(steps, action) {
   switch (action.type) {
-    case "reset":
-      return action.payload;
-
+    case "reset": {
+      return initialSteps;
```
lionelB: we should return a copy of it. using `initalSteps.slice()` or `[...initialSteps]`

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
UnbearableBear: I don't think so because the stepIndex is something internal to the wizard. So asking to the consumer of wizard to code it in the reducer would mean it could also impact the way pagination is handled. But it should not except for the initial step index. The only thing index can configure is the steps displayed

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/stepReducer.js
UnbearableBear: yep !

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/index.js
lionelB: I wonder if we can not group Rules and stepReducer since they are really inter-dependant

lionelB (commented):
just few question but nice refacto 👍

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
lionelB: in fact I would leave the stepReducer up to the user but include it in a "WizzardReducer" wich handle the setIndex something like
```js
case steps: return {...state, steps: stepReducer(state.steps, action)}
```
since reducers can be combined.

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/index.js
UnbearableBear: I like the idea ! But I need to figure out how

lionelB: no problem, we can do it later :) we can add a reminder

UnbearableBear: I'm going to create an issue

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
UnbearableBear: Ok I added this to the issue aswell, I will get back at it as soon as the planning gets more lightweight.

UnbearableBear:
Bump !

packages/code-du-travail-frontend/src/outils/IndemniteLicenciement/Wizard.js
```diff
@@ -92,10 +95,17 @@ function Wizard({
...
+    PropTypes.shape({
+      component: PropTypes.func.isRequired,
+      name: PropTypes.string.isRequired,
+      label: PropTypes.string.isRequired
+    })
+  ).isRequired,
   initialStepIndex: PropTypes.number,
-  onSubmit: PropTypes.func.isRequired
+  initialValues: PropTypes.object,
+  Rules: PropTypes.func
```
lionelB, line 108: `PropTypes.node` should it works for func and element)

UnbearableBear: It seems to dislike it. Maybe it is because it is a basic function component ?
Screenshot 2019-07-10 at 15 48 45

lionelB (approved):

PR #1049, opened 2019-07-24
ci: update docker-compose to allow deploy using tag

lionelB:
udpate docker-compose config to rely on gitlab registry for images
that way after update the docker-compose.override setting with docker-compose.override.prod.yml

we should deploy using `CDTN_HASH=0000000 docker-compose up --build` 

@douglasduteil @revolunet seems good ?

### DIFF
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -11,14 +11,12 @@ services:
   python:
     build:
       context: ./packages/code-du-travail-data
-      dockerfile: ./Dockerfile
     volumes:
       - ./packages/code-du-travail-data/:/app
 
   nlp_api:
     build:
       context: ./packages/code-du-travail-nlp
-      dockerfile: ./Dockerfile
     environment:
       - FLASK_RUN_PORT=$NLP_PORT
       - FLASK_ENV=development
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -3,14 +3,22 @@ version: "3"
 services:
   # In development, open ES port so host API can connect to it
   elasticsearch:
+    build:
+      context: ./docker/elasticsearch
     ports:
       - 9200:9200
 
   python:
+    build:
+      context: ./packages/code-du-travail-data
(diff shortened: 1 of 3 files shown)


docker-compose.override.dev.yml
```diff
@@ -3,14 +3,22 @@ version: "3"
...
   elasticsearch:
+    build:
+      context: ./docker/elasticsearch
     ports:
       - 9200:9200
 
   python:
+    build:
+      context: ./packages/code-du-travail-data
+      dockerfile: ./Dockerfile
```
ivangabriele: This line can be removed.

docker-compose.override.dev.yml
```diff
@@ -3,14 +3,22 @@ version: "3"
...
     volumes:
       - ./packages/code-du-travail-data/:/app
 
   nlp_api:
+    build:
+      context: ./packages/code-du-travail-nlp
+      dockerfile: ./Dockerfile
```
ivangabriele: Bis.

docker-compose.override.prod.yml
```diff
@@ -1,18 +1,22 @@
 version: "3"
 
 services:
-  # In production, dockerize the frontend
+  # In production, use registry image
+  elasticsearch:
+    image: registry.gitlab.factory.social.gouv.fr/socialgouv/code-du-travail-numerique/elasticsearch:$CDTN_HASH
+    ports:
```
revolunet, line 7: 9200

revolunet: Also set a défaut when CDTN_HASH not provided ?

lionelB: there is no master tag so we should provide an arbitrary one ?

revolunet: latest ?

lionelB: nor latest ;)

revolunet: i guess latest should exist by convention ?

revolunet (approved):

PR #1079, opened 2019-08-01
fix(k8s): increase nlp startup time

revolunet:
pour @ArmandGiraud


UnbearableBear (approved):

revolunet:
cc @ylascombe @Virasack

PR #1094, opened 2019-08-12
refactor(fiche_mt): move py'extraction code in js

lionelB:
translate extraction code for fiche_min-travail in js. 
create a custom slugify function that mimic slugify from python.
there is upcoming PR (code-du-travail, kali, theme) for replace the slugify package
with this new function


revolunet:
good move, but shouldnt we use an existing slugify package ?

revolunet (approved):

PR #1163, opened 2019-08-25
chore(deps): remove dependency unist-util-parents

douglasduteil:
Does `unist-util-parents` useful ?


douglasduteil:
I'm not seeing any references to `unist-util-parents` in the code...
So I'm removing it

lionelB (approved):
apparently not :) good catch !

lionelB:
@douglasduteil need a rebase

PR #1184, opened 2019-08-26
fix(gitlab): follow renovate update

revolunet:
following renovate upgrade here : https://github.com/SocialGouv/code-du-travail-numerique/pull/1129

should fix some snapshots score issues


douglasduteil (changes requested):
Oh I didn't saw that...
We should use our commun `docker/elasticsearch/Dockerfile` image here

PR #1206, opened 2019-08-28
docs(README): update api usage

douglasduteil:

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ## Installation
 
-Make sure you're using NodeJS 10.15 (9 or 11 won't work).
+Make sure you're using NodeJS 10+.
 
 ```sh
 # Install all the packages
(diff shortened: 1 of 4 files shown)


lionelB:
@douglas, why do we need to copy file on postbuild ?

douglasduteil:
I think we should remove my hacky postbuild copy. It was necessary to make the build (dist/index.js) work locally. 
I think that the priority would be to make it work in the Dockerfile and document the additional needs in the README.

What you think @lionelB ?

PS : Tag me as @douglasduteil. Not @douglas
PPS : Sorry about the noise @douglas :pray:

douglasduteil:
BTW we might want to try removing the whole hack as `ncc@0.20` and `elastic-apm-node@2.16` seems to more compatible 

Screenshot from 2019-08-29 09-41-00

lionelB:
agree to remove ! I also add to update ncc in the PR #1089

lionelB (approved):

PR #1227, opened 2019-09-04
UI preliminary work

UnbearableBear:
What's in there ?

- I added -old to the current ui in preparation for the new shiny incoming ui
- I fixed some new linting errors due to the clumsy use of some hooks
- Fixed the focus ring which was not working since a few months
- I updated the react-accessible-accordion
- I updated docz to 1.3 (nothing changed in fact)

Additional note: I also miserably failed to update docz to 2.0.0-rc


revolunet (approved):
👍

PR #1245, opened 2019-09-10
fix(nlp): docker data embed

revolunet:
Embed dump.data.json + suggester data in docker image at build time


revolunet:
il n'y a "plus que" le lancement de gunicorn qui pose un souci

ArmandGiraud (approved):

PR #1246, opened 2019-09-11
New UI

UnbearableBear:
La nouvelle UI est là !
Seul le composant bouton a été fait pour le moment.

Screenshot 2019-09-11 at 16 25 13
---------------------------------------
Screenshot 2019-09-11 at 16 25 22

J'ai aussi rajouté une règle à eslint concernant les packages afin qu'il rale lorsque des paquets non déclarés dans le packages.json sont importés et fonctionnent quand même grace au hoisting de yarn. Ca évitera des problèmes à l'avenir si on décide de split les packages pour de bon.


package.json
```diff
@@ -30,11 +30,10 @@
...
     "@babel/preset-env": "7.5.5"
   },
-  "workspaces": [
-    "packages/*",
-    "packages/code-du-travail-data/dataset/!(fiches_service_public)",
-    "packages/code-du-travail-data/dataset/code_du_travail/eposeidon_script"
-  ],
+  "workspaces": {
+    "packages": ["packages/*", "packages/code-du-travail-data/dataset/*"],
+    "nohoist": ["**/react-ui", "**/react-ui/**"]
```
UnbearableBear, line 35: Impératif sinon docz crashe lors de son build car il ne va pas chercher react & co à la racine du proj
(pour le moment en tous cas)

packages/code-du-travail-data/dataset/fiches_service_public/package.json
```diff
@@ -22,7 +22,6 @@
     "fetchData": "node fetchData.js",
     "start": "yarn backup && node index.js > fiches-sp-travail.json",
     "report": "node diff.js && node missingTheme.js",
-    "poststart": "yarn updateTheme && yarn report",
-    "test": "jest"
```
UnbearableBear, line 26: Il n'y a pas de tests dans ce repo

packages/code-du-travail-frontend/package.json
```diff
@@ -16,20 +16,22 @@
...
     "debounce-promise": "^3.1.2",
+    "express": "^4.17.1",
     "express-sitemap-xml": "^1.1.0",
     "final-form": "^4.18.5",
     "final-form-arrays": "^3.0.1",
     "final-form-calculate": "^1.3.1",
     "isomorphic-unfetch": "^3.0.0",
     "memoizee": "^0.4.14",
     "next": "^9.0.5",
+    "prop-types": "^15.7.2",
```
lionelB, line 34: this is a dev-dep

packages/code-du-travail-frontend/package.json
```diff
@@ -50,14 +52,15 @@
     "@babel/core": "^7.6.0",
     "@babel/preset-env": "^7.6.0",
     "@socialgouv/eslint-config-react": "^0.9.6",
+    "@testing-library/react": "^9.1.4",
```
lionelB: dev-dep too

packages/react-ui/rollup.config.js
```diff
@@ -0,0 +1,32 @@
...
+      format: "esm"
+    }
+  ],
+  external: ["prop-types", "react", "react-dom", "styled-components"],
+  plugins: [
+    babel(),
+    resolve({
+      mainFields: ["module", "jsnext", "jsnext:main", "main"]
+    }),
+    commonjs(),
```
lionelB, line 24: I thinks we can remove the commonjs() plugin here

packages/react-ui/src/icons/Cross.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
```
lionelB, line 1: we don't use react-feather anymore ?

lionelB (commented):
we would prefer a name tied to the project rather than react-ui :)
also, all the Theme related code to allow theme swapping should be add when we do a dark mode ;)

packages/react-ui/src/icons/Cross.js
UnbearableBear: I was supposed to start a new ui from scratch and they were no react-feather icons in the new design, I had all the SVG provided.
But we won't be doing the new design so I guess there is no need to spend too much time on this review.
Still you can use this branch to potentially play with rebass. You might be more successful implementing it in a fashion you like !

UnbearableBear:
I think the only reason to use this new ui is the theming now that we dropped the new design. So I guess that removing theming code make this new ui concept deprecated for now. But I'm ok with keeping the old ui and close the two PR

packages/react-ui/src/icons/Cross.js
lionelB: Ok, is it possible to extract the eslint addition and fix you do ?

UnbearableBear: Absolutely !

UnbearableBear: I'll let the big PR be merged first the I'll take it.

PR #1269, opened 2019-09-16
feat(glossaire): add glossary page

lionelB:
This PR add glossary page and definition 
need to add the snippet on content

related to #1186


packages/code-du-travail-data/dataset/glossaire/index.js
```diff
@@ -1,36 +1,22 @@
...
+  const response = await fetch(RECORDS_URL);
+  const items = await response.json();
+  return items.data.map(({ title, abbrs, variants, definition, refs }) => ({
+    title,
+    slug: slugify(title),
+    abbrs,
+    variants,
+    definition,
+    refs
+  }));
```
revolunet: `abbrs` and `variants` are multiline string, we should split on `\n` on fetch ?

revolunet:
shouldnt we put that in `datafiller/glossaire.data.json` (we have themes and prequalified there and maybe more later)

packages/code-du-travail-data/dataset/glossaire/index.js
lionelB: Alright!  we should translate to an array before json

revolunet (approved):

lionelB:
j'attend les retours de @virginielastisneres  et on merge, on fera les tooltip dans une deuxieme PR

lionelB:
@revolunet bon finalement les tooltip sont aussi là 😅

PR #1324, opened 2019-09-25
fix(nlp): update nlp docker config

lionelB:
add logs and replace workers with threads in nlp_api, also fix some typo in docker example


revolunet:
a merger avec #1330 ?

revolunet (approved):

PR #1333, opened 2019-09-26
chore(CI): add Dockerlint with hadolint

revolunet:
Add [hadolint](https://hadolint.github.io/hadolint) for all Dockerfiles


lionelB:
47 checks... not bad :)

packages/code-du-travail-nlp/Dockerfile
```diff
@@ -24,7 +26,11 @@ ENV FLASK_APP api
 ENV FLASK_APP="development"
 
 RUN python -m venv venv
+
+# hadolint ignore=SC1091
```
lionelB, line 30: what does it means ? 
should we fix that in a near future ?

lionelB (approved):
HADOKEN!

packages/code-du-travail-nlp/Dockerfile
revolunet: this happens when the linter cant find the mentioned source file

im not sure we can "fix" it 🤔 

https://github.com/koalaman/shellcheck/wiki/SC1091

PR #1355, opened 2019-09-30
feat(docker): use limited user privileges for #1307

revolunet:
api and frontend containers : use unprivilegied user and  port

### DIFF
--- a/.gitlab-ci/stages/deploy.yml
+++ b/.gitlab-ci/stages/deploy.yml
@@ -44,7 +44,7 @@ include:
     CONTEXT: api
     ELASTICSEARCH_LOG_LEVEL: trace
     IMAGE_NAME: ${CI_REGISTRY_IMAGE}/api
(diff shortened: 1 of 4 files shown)


.gitlab-ci/stages/deploy.yml
```diff
@@ -44,7 +44,7 @@ include:
     CONTEXT: api
     ELASTICSEARCH_LOG_LEVEL: trace
     IMAGE_NAME: ${CI_REGISTRY_IMAGE}/api
-    PORT: 80
+    PORT: 8000
```
lionelB, line 47: why there is no ingress update too ?

lionelB (approved):

.gitlab-ci/stages/deploy.yml
revolunet: looks like the api ingress rule also use this variable

PR #1409, opened 2019-10-07
fix(CI): add ENVIRONMENT=production

revolunet:
to fix https://github.com/SocialGouv/code-du-travail-numerique/blob/2faf9588f4203826c46a487652e07a3088c22741/packages/code-du-travail-frontend/server/server.js#L10


lionelB (approved):
sounds good

PR #1480, opened 2019-10-16
change suggest URL to use remote lucene based API

rmelisson:


lionelB:
Can we close it in favor of #1549

rmelisson:
deprecated

PR #1525, opened 2019-10-21
fix(precarite): fix isNotYetProcessed message

### LINKED ISSUE #1521: [Bug] Outil Indemnité de précarité
**Rappel**
Lorsque l'usager sélectionne une convention collective non traitée encore dans le fichier excel, s'affiche un disclaimer indiquant que sa CC n'est pas encore traitée mais que nous l'invitons à continuer la simulation. 

**Description du bug**
Ne s'affiche plus le disclaimer alertant les usagers ayant sélectionner une CC non traitée... 

**Screenshots**
image
image

lionelB:
fix #1521


UnbearableBear (approved):
Not a big fan of the naming since I'm not sure what should be processed when you type in "toto" but LGTM !

lionelB:
@UnbearableBear `isCCnHandled` is better ?  if so, @nkmr can do the renaming since he is currently refactoring this part

PR #1549, opened 2019-10-23
proto elastic suggester

lionelB:
wip


rmelisson:
Fix #1540 #1542 #841

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
```diff
@@ -0,0 +1,39 @@
...
+test("return suggestions for re in the right format", () =>
+  getSuggestions("re")
+    .expect(200)
+    .expect("Content-Type", /json/)
+    .expect(res => expect(res.body).toMatchSnapshot()));
+
+test("accentuation is ignored", async () =>
+  getSuggestions("ré").expect(res =>
+    // this is ugly, coulnt find a better way though
+    expect(res.body).toEqual(expect.arrayContaining(["retraite"]))
```
UnbearableBear: I would write it in natural javascript `expect(res.body.includes("retraite")).toBeTruthy()`. Does that looks better to you ?

UnbearableBear: Maybe this would work too: `expect(res.body).toContain("retraite")`

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+test("return suggestions for re in the right format", () =>
+  getSuggestions("re")
+    .expect(200)
+    .expect("Content-Type", /json/)
+    .expect(res => expect(res.body).toMatchSnapshot()));
+
+test("accentuation is ignored", async () =>
+  getSuggestions("ré").expect(res =>
```
UnbearableBear: you could also use the async await mecanism here in order to avoid having multiple `.expect`but not 100% sure.

packages/code-du-travail-api/src/server/routes/suggest/index.js
```diff
@@ -0,0 +1,36 @@
...
+const minQueryLength = 2;
+const suggestionsSize = 5;
+
+/**
+ * Return the search suggestion
+ *
+ * @example
+ * http://localhost:1337/api/v1/suggest?q=aba
+ *
+ * @returns {Object} An object containing the matching theme .
```
UnbearableBear: outdated comment 🌵

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
lionelB: you could first use await (you already added the `async`)
```js
test("accentuation is ignored", async () => { 
  const response = await getSuggestions("ré")
  expect(response.body.include("retraite")).toBeTruthy()
  // see https://jestjs.io/docs/en/using-matchers.html#arrays-and-iterables
  expect(res.body).toContain("retraite"
}
```

rmelisson:
whoops, my bad c'est toujours WIP, je crois pas que je puisse revert de open à draft

packages/code-du-travail-nlp/CHANGELOG.md
```diff
@@ -3,138 +3,81 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-## [3.3.1](https://github.com/SocialGouv/code-du-travail-numerique/compare/v3.3.0...v3.3.1) (2019-10-08)
-
-**Note:** Version bump only for package @cdt/nlp
+## [Unreleased]
```
lionelB: You don't need to update the changelog, it will be write automatically

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
rmelisson: One expect comes from supertest and the other one is from jest, no ?
I changed this anyway according to your other comment, feel free to resolve if it looks good to you.

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function ensureSuggestionsMatchSnapshot(query) {
+  return getSuggestions(query).expect(res =>
+    expect(res.body).toMatchSnapshot()
+  );
+}
+
+test("return suggestions for re in the right format", () =>
+  getSuggestions("re")
```
lionelB: can you use async in test case ? It would be easier to read.
 ```js 

test("return suggestions for re in the right format", async () => {
  const response = await request(app.callback()).get(`/api/v1/suggest?q=re`);
  expect(response.status).toBe(200)
  expect(response.get("Content-type").toMatch(/json/);
  expect(response.body).toMatchSnapshot()
})
```

if can also create a getSuggestion helder
```js
async function getSuggestion(query) {
   return request(app.callback()).get(`/api/v1/suggest?q=${query}`);
}
```

packages/code-du-travail-data/Dockerfile
```diff
@@ -12,6 +12,8 @@ FROM ${NLP_IMAGE} as cdtn-nlp-image
 
 FROM node:10-alpine
 
+RUN apk add --no-cache curl=7.64.0-r3 && rm -rf /var/cache/apk/*
```
lionelB: `rm -rf /var/cache/apk/*` is not needed  think

packages/code-du-travail-data/indexing/suggestion.js
```diff
@@ -0,0 +1,95 @@
...
+  const client = new Client({
+    node: `${ELASTICSEARCH_URL}`
+  });
+
+  const ts = Date.now();
+  const tmpIndexName = `${SUGGEST_INDEX_NAME}-${ts}`;
+
+  await populateSuggestions(client, tmpIndexName);
+
+  await client.indices.putAlias({
```
lionelB: you should better use the `updateAliases`  like you did in the [indexing/index.js](https://github.com/SocialGouv/code-du-travail-numerique/pull/1549/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.py&file-filters%5B%5D=.txt&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&file-filters%5B%5D=dotfile#diff-de7bcad4c0f097647e9c4c8cc8f6bf1dL104-L110)

lionelB:
@rmelisson Looks good. just few comments :)  nice job!

packages/code-du-travail-api/src/server/routes/__tests__/suggest.spec.js
rmelisson: `getSuggestions` still return a promise, but the test cases are now using async for readability. Hope you're fine with this.

rmelisson (approved):

PR #1580, opened 2019-10-28
feat: scan docker containers with snyk

revolunet:
ajoute un scan de vuln snyk sur les images docker avec allow_failure=True

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,19 +1,21 @@
 ---
 include:
-- /.gitlab-ci/variables.yml
-- /.gitlab-ci/extends.yml
+  - /.gitlab-ci/variables.yml
+  - /.gitlab-ci/extends.yml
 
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -1,19 +1,21 @@
...
-- "Prepare"
-- "Code Quality"
-- "Registration"
-- "Data"
-- "Deploy"
-- "Notify Finished Deployment"
+  - "Prepare"
+  - "Code Quality"
+  - "Registration"
+  - "Vuln scanner"
```
lionelB, line 18: VulScanner should be after data since it also a registration stage

revolunet: data is not exposed so not sure this is really useful

lionelB: oki !

revolunet:
approved ?

lionelB (approved):

PR #1653, opened 2019-11-06
Armand giraud/sem search update

ArmandGiraud:
- change the preprocessing & use a common function
- refactor a little

⚠️ the preprocessing has a big impact on relevance
some positive and some negatives but global relevance increase from 29.7 to 31.2

check http://c9f2a2f.code-du-travail-numerique.dev.factory.social.gouv.fr/ for tests

ex: 
image
image

it is possible to get a changelog on datafiller

### DIFF
--- a/packages/code-du-travail-nlp/api/sem_search.py
+++ b/packages/code-du-travail-nlp/api/sem_search.py
@@ -1,10 +1,10 @@
+"""Encode a query string or a document with a dual encoder"""
 
 import logging
 import json
 import tensorflow as tf
 import tensorflow_hub as hub
 import tf_sentencepiece
-import unicodedata
 import time
 
 stops_path = "../data/stops.txt"
@@ -12,11 +12,6 @@ stops_path = "../data/stops.txt"
 cache_path = "~/Downloads"
 
 logger = logging.getLogger("gunicorn.errror")
-
-
-def add_slash(c):
-    c["slug"] = f"/{c['slug']}"
-    return c
 
 
 class SemSearch():
@@ -27,6 +22,9 @@ class SemSearch():
 
         with open(stops_path, "r") as f:
             stops = f.read().splitlines()
+            stops.extend(["travail",
+                          "travailler",
+                          "travaille"])
 
         # make a hashtable for performance
         strp = self.strip_accents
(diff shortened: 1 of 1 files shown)


packages/code-du-travail-nlp/api/sem_search.py
```diff
@@ -27,17 +22,19 @@ def __init__(self, stops_path: str):
 
         with open(stops_path, "r") as f:
             stops = f.read().splitlines()
+            stops.extend(["travail",
+                          "travailler",
+                          "travaille"])
```
revolunet, line 27: why not in the data file ?

revolunet:
i see some changes in the unicode handling; no impact on accentuated requests ?

packages/code-du-travail-nlp/api/sem_search.py
ArmandGiraud: because the data file is currently used by both the sem search and the suggester and removing "travail" from suggestions is a bad idea I guess 😄 .
If we move suggester to elastic, we might want to use two different files

ArmandGiraud:
There will be an impact on:
- forgotten accents by the user (negative)
- correct accents set by the user (positive)

we might want to remove wrongly accentuated requests from the suggestions...
@ClementChapalain any recommandation on this?

ClementChapalain:
I'd need some more explanations on this please : 
- forgotten accents by the user (negative)
- correct accents set by the user (positive)

any examples :) ?

ArmandGiraud:
je te mets 2 examples pour l'évolution:

## Impact négatif:
### Avant Evolution

image

### Après
image

## impact positif:

### Avant:
image

### Après:

image

ClementChapalain:
j'ai l'impression que le premier cas (l'accent est oublié) est plus fréquent que le second (l'utilisateur utilise la bonne accentuation) et que donc on perd en qualité, mais c'est juste mon avis sur ce point particulier, je n'ai pas la vue sur toute la PR

packages/code-du-travail-nlp/api/sem_search.py
lionelB: suggester is gone, stopwords in no more shared ! we can update the stop_words list (and retrieve only the stopwords file)

lionelB (approved):

lionelB:
@ArmandGiraud j'ai supprimé l'env. pour en repopper un le mieux serait de rebaser la PR (vu que le suggester n'existe plus)

revolunet:
closable ?

revolunet:
i guess we can close this one

PR #1659, opened 2019-11-07
chore: more error reporting

revolunet:
better errors in indexing. `logger.error` only accept strings

### DIFF
--- a/packages/code-du-travail-data/indexing/index.js
+++ b/packages/code-du-travail-data/indexing/index.js
@@ -130,7 +130,9 @@ async function main() {
 
 main().catch(response => {
   if (response.body) {
-    logger.error(response.body.error.reason || response.body);
+    logger.error(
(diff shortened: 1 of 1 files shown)


lionelB (approved):

packages/code-du-travail-data/indexing/index.js
```diff
@@ -130,9 +130,9 @@ async function main() {
 
 main().catch(response => {
   if (response.body) {
-    logger.error(response.body.error.reason);
+    logger.error(response.body.error.reason || response.body);
```
lionelB:
```suggestion
    logger.error(response.body.error && response.body.error.reason  || response.body);
```

lionelB: @revolunet wdyt ?

revolunet:
wow, GH suggestions are awesome, thank thats even better

PR #1669, opened 2019-11-08
fix(piwik): allow duplicate trackings

revolunet:
consecutive searchs were not tracked, this fix it and should give more correct visitors logs


UnbearableBear (approved):

ArmandGiraud:
Nice, could we also add another custom event like [SelectedSuggestion](https://github.com/SocialGouv/code-du-travail-numerique/blob/98de1a9f7788484557ada309bd37d341c205197f/packages/code-du-travail-frontend/src/search/SearchBar/index.js#L42) to track the event click on "retour aux résultats". It would avoid confusion in the parsing of logs?
image

PR #1699, opened 2019-11-13
feat(courrier-types): remove manual script step on docs update

revolunet:
remote the necessary `yarn start` when adding some model. Make it automatically at build time


packages/code-du-travail-data/dataset/courrier-type/package.json
```diff
@@ -8,7 +8,7 @@
     "doc": "docs"
   },
   "scripts": {
-    "start": "node index > ../export-courriers.json"
+    "start": "node index > ./export-courriers.json"
```
lionelB, line 11: 👍

lionelB (approved):

PR #1708, opened 2019-11-14
feat(matomo): add suggestionCandidate and resultsCandidate in matomo

lionelB:
better tracking of provided info to user
- send candidateResults when display a SERP (Search Engine Result Page)
- send candidateSuggestion when clicking a suggestion.
- handle campaign id
- log back to results
- log selectResult 

this will help us better know what suggestions are better clicked 

ref #1624

### DIFF
--- a/packages/code-du-travail-frontend/__mocks__/next/router.js
+++ b/packages/code-du-travail-frontend/__mocks__/next/router.js
@@ -1,25 +1,28 @@
+const route = jest.fn();
+const router = {
+  asPath: "mock",
+  route,
+  pathname: "mock",
+  query: { q: undefined },
+  push: jest.fn(path => {
+    route(path);
+    return Promise.resolve();
+  })
+};
+
 module.exports = {
+  router,
+  mocked: true,
   events: {
     on() {},
     off() {},
     trigger() {}
   },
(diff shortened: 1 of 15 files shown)


revolunet:
Not sure this is a good idea to push every results and suggestions to the user and back to the server; cant we extract these at "logs pipeline" level ?

lionelB:
> cant we extract these at "logs pipeline" level ?

that's what was done but its also ended with weird results when trying to consolidate those logs. 
@rmelisson and @ArmandGiraud ask for it, so I leave them to better explain the why :)

ArmandGiraud:
> cant we extract these at "logs pipeline" level ?

We can, but sending the suggestions and result list directly in Matomo avoids errors in the logs pipeline. Secondly it avoids to make all requests twice, as it is not necessary 🌎.
@revolunet  Why do you think it's not a good idea to send it?

revolunet:
it just sounds weird to send the results to the user and then send them back to us... also, network and performance from a user perspective :) also as a user i wont be very happy to see this in my network logs...
BTW we need reliability on the data so if we cant reproduce the results on our side we dont have much choice....

revolunet (approved):

PR #1709, opened 2019-11-14
feat(api): add score and algo key in search results

lionelB:
add `_score` and `algo` keys in search results
algo could one of `["pre-qualified", "semantic", "fulltext"]`

cc @ArmandGiraud @rmelisson


revolunet (approved):

ArmandGiraud (approved):
perfect 👌

PR #1760, opened 2019-11-20
feat(frontend): new-ui-setup

UnbearableBear:
An ugly, yet necessary, step toward the new design
Screenshot 2019-11-21 at 15 57 55

Screenshot 2019-11-21 at 16 02 04
Screenshot 2019-11-21 at 16 02 46
Screenshot 2019-11-21 at 16 04 00

Review advice: 
Pull the branch, turn the Storybook on with `yarn start` and go check the changes


packages/code-du-travail-frontend/pages/_document.js
```diff
@@ -42,27 +42,17 @@ export default class MyDocument extends Document {
     return (
       <html lang="fr" prefix="og: http://ogp.me/ns#">
         <Head>
-          <link rel="stylesheet" type="text/css" href="/static/fonts.css" />
+          <link
+            href="https://fonts.googleapis.com/css?family=Merriweather:400,700|Open+Sans:400,600i&display=swap"
```
lionelB, line 46: I think we should host our fonts (poke @revolunet )

revolunet: yes we have to :)

UnbearableBear: We can inline them (like it was before). But it is more convenient like that for now. Inlining the fonts increases the bundle size but remove the "font swap" you can see during the first load.
I have a preference for the "swap" but it is visually less elegant than inlining the fonts. And yes once the design is finished, I'll provide you with the exact set needed so you can host it if you don't wish to inline them.

revolunet: yes we have to host them or embed them directly :/

lionelB: we probably want to subset them to :)

lionelB (approved):
j'ai pas review on iterera ensuite si y'a besoin de clean 
edit: GOooooood job !

PR #1812, opened 2019-11-25
chore: the almighty cdtn cleaner

lionelB:
still wip since the script cannot be launch right now. but you can already review it.
remove merged feature branch k8s object and es index


lionelB:
closed in favor of https://github.com/SocialGouv/docker/pull/64

PR #1815, opened 2019-11-26
(feat) fuzzy matching for datafiller queries

rmelisson:
PR for #1731 

Using fuzzyball for now, should do the trick for now : fuzzy matching with simple ratio > 90%
https://www.npmjs.com/package/fuzzball#usage

For more advanced approaches I'd suggest to take time to evaluate a more global strategy that would involve ES.

With a bit of optimization the overhead in term of query time is ~10/15ms, this is a bit critical as the datafiller matching is evaluated for *every* query. I've included lodash.deburr in order to help with preprocessing (mostly removing accentuation) and reduce matching time (~100ms).


packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -2,12 +2,61 @@ const knownQueries = require("@cdt/data...datafiller/prequalified.data.json");
...
+  // preprocess query
+  const ppQuery = preprocess(query);
+
+  const results = fuzz.extract(
+    ppQuery,
+    Object.keys(knownQueriesSet),
+    fuzzOptions
+  );
+
+  const closerMatch = results[0][0];
```
revolunet: we should ensure we have at least one result here ?

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -2,12 +2,61 @@ const knownQueries = require("@cdt/data...datafiller/prequalified.data.json");
...
+  const bestScore = results[0][1];
+
+  if (bestScore > threshold) {
+    return knownQueriesSet[closerMatch];
+  } else {
+    return undefined;
```
revolunet: (return undefined is the default when the function returns nothing)

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -17,4 +66,6 @@ const getSavedResult = async query => {
   }
 };
 
+populateSavedQueries();
```
revolunet: maybe `const knownQueriesSet = populateSavedQueries();` ?

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
rmelisson: yeah that's what I figured, but thought return was more explicit, I'll remove it

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
rmelisson: we always get results, unless there are no savedQueries, but I can add the test

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -2,12 +2,61 @@ const knownQueries = require("@cdt/data...datafiller/prequalified.data.json");
...
+// Preprocess query : remove accentuation and
+//  non alpha numerical to speed up the lookup.
+const preprocess = q => {
+  return fuzz.full_process(deburr(q.trim()));
+};
+
+// We populate the saved queries in an object in order to ease
+//  searches and map variant matches with the actual known query.
+const populateSavedQueries = () => {
+  knownQueries.map(q => {
```
lionelB: use forEach or a reduce function here
```js
const knownQueries = knowQueries.reduce((queries, querie) => {
  for (const variant of querie.variants) {
    const prepro = preprocess(variant);
    queries[prepro] = querie
  }
  return queries
}
```

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -2,12 +2,61 @@ const knownQueries = require("@cdt/data...datafiller/prequalified.data.json");
...
+      const prepro = preprocess(v);
+      knownQueriesSet[prepro] = q;
+    });
+  });
+};
+
+// Test if a given query fuzzy matches with
+//  a known one (and variants).
+const testMatch = query => {
+  const fuzzOptions = {
```
lionelB: this could be extract from the function

packages/code-du-travail-api/src/server/routes/search/search.getSavedResult.js
```diff
@@ -2,12 +2,61 @@ const knownQueries = require("@cdt/data...datafiller/prequalified.data.json");
...
+    unsorted: false,
+    limit: 2
+  };
+
+  // preprocess query
+  const ppQuery = preprocess(query);
+
+  const results = fuzz.extract(
+    ppQuery,
+    Object.keys(knownQueriesSet),
```
lionelB: this could be extract from the function too

lionelB (approved):

revolunet (approved):

PR #1821, opened 2019-11-27
(feat) precise algo type in search log events

rmelisson:
add algo type to select_result and use "both" algo when fulltext+sem are matching

### DIFF
--- a/packages/code-du-travail-frontend/src/search/SearchResults/index.js
+++ b/packages/code-du-travail-frontend/src/search/SearchResults/index.js
@@ -13,11 +13,16 @@ const SearchResults = ({
 }) => {
   const router = useRouter();
   useEffect(() => {
-    const toSlug = ({ source, slug }) => `${source}/${slug}`;
+    const toLogCandidate = ({ source, slug, algo }) => {
+      return {
+        slug: `${source}/${slug}`,
+        algo
+      };
+    };
     const results = JSON.stringify({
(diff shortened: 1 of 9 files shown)


packages/code-du-travail-frontend/src/search/SearchResults/index.js
```diff
@@ -13,11 +13,16 @@ const SearchResults = ({
 }) => {
   const router = useRouter();
   useEffect(() => {
-    const toSlug = ({ source, slug }) => `${source}/${slug}`;
+    const toLogCandidate = ({ source, slug, algo }) => {
+      return {
+        slug: `${source}/${slug}`,
```
lionelB: should we use getRouteBySource here (or remove it from upper code ?)

rmelisson: Ah yes good find, for now the urls we send are different, I was making the match in the logs analysis, but that should be done properly here. 
I guess this is the recommended way to format it ?
```
const trackedUrl = 
   source === SOURCES.EXTERNALS ? url : `/${getRouteBySource(source)}/${slug}`;
```

lionelB: yep !

rmelisson: done

lionelB (approved):

PR #1888, opened 2019-12-05
Fixed tools situations reset fields

nkrmr:
Fixed reset fields when parent changed

### DIFF
--- a/packages/code-du-travail-frontend/src/outils/common/situations.utils.js
+++ b/packages/code-du-travail-frontend/src/outils/common/situations.utils.js
@@ -112,3 +112,14 @@ export function getRef({ ref, refUrl }) {
     </p>
   );
 }
+
+export const getFormProps = (key, initialSituations, pastQuestions) =>
+  // list keys that no longer exist
+  Object.keys(initialSituations)
+    .filter(k => !pastQuestions.find(([key]) => k === key))
+    .concat(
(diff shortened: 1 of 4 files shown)


packages/code-du-travail-frontend/src/outils/common/situations.utils.js
```diff
@@ -112,3 +112,23 @@ export function getRef({ ref, refUrl }) {
...
+    // list keys that no longer exist
+    const resetFormProps = Object.keys(initialSituations)
+      .filter(k => !pastQuestions.find(([key]) => k === key))
+      .concat(
+        // list keys that need to be reseted
+        pastQuestions
+          .slice(pastQuestions.findIndex(([k]) => k === key) + 1)
+          .map(([key]) => key)
+      );
+    resetFormProps.forEach(key => form.change(`criteria.${key}`, undefined));
```
lionelB: `criteria` should be extracted unless we want to forget to update it is forms data change ;)

lionelB (changes requested):
seems alsmost good. just missing a small test

packages/code-du-travail-frontend/src/outils/common/situations.utils.js
nkrmr: Of course

lionelB (approved):
👌

PR #1916, opened 2019-12-09
feat: basic website usable without API

revolunet:
Make homepages, tools and static page work without any API

### DIFF
--- a/packages/code-du-travail-frontend/pages/index.js
+++ b/packages/code-du-travail-frontend/pages/index.js
@@ -1,6 +1,7 @@
 import React from "react";
 import getConfig from "next/config";
 import fetch from "isomorphic-unfetch";
+import * as Sentry from "@sentry/browser";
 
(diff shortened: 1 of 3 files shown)


packages/code-du-travail-frontend/pages/index.js
```diff
@@ -27,12 +28,18 @@ const Home = ({ pageUrl, ogImage, children = [] }) => (
 );
 
 Home.getInitialProps = async () => {
-  const response = await fetch(`${API_URL}/themes`);
-  if (!response.ok) {
-    return { statusCode: response.status };
+  try {
```
lionelB, line 31: should this be done in _app ?

revolunet: the thing is we dont have the same props shape

UnbearableBear:
Merged on my own branch to avoid conflicts, PR will come later

PR #1943, opened 2019-12-11
fix: courriers indexation

revolunet:
- cleanup du fichier courriers.json
 - remove need of manual "build"
 - html is directly extracted at indexation time

same coming for fiches sp


lionelB (commented):
should we add it to .gitignore

revolunet:
the file is removed forever so no need to ignore it ?

lionelB:
each time someone launch `yarn build` it will generate it.

revolunet:
non en fait là ce fichier intermediaire n'est plus généré. l'extract HTML est réalisé au moment de l'indexation

packages/code-du-travail-data/indexing/populate.js
```diff
@@ -136,30 +137,30 @@ function* cdtnDocumentsGen() {
...
-      description,
-      text: questions.join("\n"),
-      html,
-      filename,
-      date: date_redaction,
-      editor: source,
-      author: redacteur,
-      tags: flattenTags(tags)
-    })
+  yield getCourriers().then(courriers =>
```
lionelB: yielding a promise unwrap it ? 
o_O  didn't know we could do that :)

revolunet: no you cant :)

revolunet:
up :)

packages/code-du-travail-data/indexing/slug_checker.js
```diff
@@ -1,15 +1,19 @@
 import { cdtnDocumentsGen, getDuplicateSlugs } from "./populate";
 import { logger } from "./logger";
 
-const documents = cdtnDocumentsGen();
-const duplicateSlugs = getDuplicateSlugs(documents);
+const checkDuplicates = async () => {
+  const documents = await cdtnDocumentsGen();
+  const duplicateSlugs = getDuplicateSlugs(documents);
```
lionelB, line 6: Does `await` missing here ?

lionelB (approved):

revolunet:
thank a ton !

PR #1989, opened 2019-12-16
refactor(front): tool layout and mobile

UnbearableBear:
Je vous invite à regarder la branche déployée


packages/code-du-travail-frontend/src/layout/css/Headings.js
```diff
@@ -18,15 +18,31 @@ export default createGlobalStyle`
     font-weight: normal;
     font-size: ${fonts.sizes.headings.large};
     font-family: 'Merriweather', serif;
+    @media (max-width: ${breakpoints.mobile}) {
```
lionelB, line 20: Since we have heading, shouldn't we remove that globalStyles ( or only add it in fichesMT? ) and add adpative styles to headings to ?

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
```diff
@@ -46,18 +48,24 @@ PrevNextBar.defaultProps = {
 
 export { PrevNextBar };
 
-const { spacings } = theme;
+const { breakpoints, spacings } = theme;
+
+const StyledButton = styled(Button)``;
```
lionelB: unused object

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
```diff
@@ -46,18 +48,24 @@ PrevNextBar.defaultProps = {
...
 
 const Box = styled.div`
   display: flex;
+  flex-direction: row-reverse;
```
lionelB, line 61: event I would naturally done the opposite, move the first primary action first in the dom is a good idea !

packages/code-du-travail-frontend/src/outils/common/StepList.js
```diff
@@ -0,0 +1,112 @@
...
+import styled from "styled-components";
+import { FlatList, theme, Wrapper } from "@socialgouv/react-ui";
+
+export const STEP_LIST_WIDTH = "28rem";
+
+export function StepList({ activeIndex = 0, items = [] }) {
+  return (
+    <StyledWrapper variant="dark">
+      <Title>
+        Étape<Plural>s</Plural>
```
lionelB: Plural is misleading, I would prefer `<HideOnMobile /> `

packages/code-du-travail-frontend/src/outils/common/StepList.js
```diff
@@ -0,0 +1,112 @@
...
+        ))}
+      </FlatList>
+    </StyledWrapper>
+  );
+}
+
+const { spacings, breakpoints } = theme;
+
+const StyledWrapper = styled(Wrapper)`
+  position: absolute;
```
lionelB, line 37: why do you need position absolute here ? this looks more complicated than before

packages/react-ui/.storybook/preview-head.html
```diff
@@ -22,6 +22,9 @@
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
     text-rendering: optimizeLegibility;
+    @media (max-width: 600px) {
```
lionelB: Could we add global styles to react-ui and use globalStyle definition in `@cdt/frontend` ?

lionelB (dismissed):

packages/code-du-travail-frontend/src/layout/css/Headings.js
UnbearableBear: The main thing is that heading components have stripes (altough we could have add option to render them without stripes) but then why should we use a component ? Plus as you say they still are raw sources of content. From fiches MT to contributions, the amount of work to get to this right would probably require several pull requests. Maybe when we'll have more time

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
UnbearableBear: This one is for mobile but I'll do it in a different way. (I'm doing a `${StyledButton} + ${StyledButton} {` in a component below but I'll do it here.

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
UnbearableBear: I mostly did this for mobile. The main action is the first you reach when scrolling on mobile

packages/code-du-travail-frontend/src/outils/common/StepList.js
UnbearableBear: It is slightly more complicated but I found no other ways to make it work responsively otherwise. The absolute + padding list is quite convenient though because it allows me to respect the full width background on figma even though the container has paddings. This is one is touchy, but if you take a look at the mobile / desktop version it should start to make sense

packages/code-du-travail-frontend/src/outils/common/StepList.js
UnbearableBear: I guess you tried to add and html tag without the comment here :D Does `<S>` is ok for you ?

packages/react-ui/.storybook/preview-head.html
UnbearableBear: That's a great idea 😲  Great great idea ! I'm gonna give it a try

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
lionelB: (yeah when I say the opposite is to leave previous dom order and use row revers on mobile but these way is better

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
lionelB: Not sure I but you could do  
```js
${Button} + ${Button} {
  margin-top: ${spacings.base};
}
```

packages/code-du-travail-frontend/src/outils/common/StepList.js
lionelB: I updateed the comment, I've forget to escape it  `<HideOnMobile />`

packages/code-du-travail-frontend/src/outils/common/StepList.js
lionelB: I did, but I will do it, I probably miss something

packages/code-du-travail-frontend/src/outils/common/PrevNextBar.js
UnbearableBear: I hoped that too but because of the forward ref wrapper it's a no go :/

packages/code-du-travail-frontend/src/outils/common/StepList.js
UnbearableBear: We can't get from a column display (mobile) to a mixed up layout (the way it is on desktop) with traditionnal flexboxes and grid is not ok on ie11

UnbearableBear: https://photos.app.goo.gl/2KmB1bno2vG4MvNG6

lionelB: thanks for the clarification !

packages/react-ui/.storybook/preview-head.html
nkrmr: Yes i already did it on a project

nkrmr (approved):

packages/react-ui/.storybook/preview-head.html
UnbearableBear: Done, it works !

PR #1993, opened 2019-12-17
feat: index fiches SP directly

revolunet:
Dont keep an intermediate fiches-sp.json file

This removes the need to regenerate the file and update themes


UnbearableBear (approved):
Je suis jaloux des -33,422 lignes de code !

PR #2003, opened 2019-12-18
feat(frontend): add color switch

### LINKED ISSUE #1858: Possibilité de changer les couleurs pour avoir un meilleur contraste
Ajouter un lien "Accessibilité" dans le menu qui ouvre un menu déroulant : 
"Contraste : 
- standard
- renforcé"

Le click sur renforcé change les couleurs du site

UnbearableBear:
fix #1858
Cast an eye on the deployed branch
Screenshot 2019-12-18 at 14 28 04

### DIFF
--- a/packages/code-du-travail-frontend/pages/_app.js
+++ b/packages/code-du-travail-frontend/pages/_app.js
@@ -2,15 +2,17 @@ import React from "react";
 import PropTypes from "prop-types";
 import App from "next/app";
 import getConfig from "next/config";
-import { ThemeProvider } from "styled-components";
 import * as Sentry from "@sentry/browser";
-import { GlobalStyles, theme } from "@socialgouv/react-ui";
+import { GlobalStyles } from "@socialgouv/react-ui";
+
 import ErrorPage from "./_error";
 
 import CookieConsent from "../src/common/CookieConsent";
 
 import { initPiwik } from "../src/piwik";
 import { initializeSentry } from "../src/sentry";
+
+import { ThemeProvider } from "../src/layout/ThemeProvider.js";
 
 const {
   publicRuntimeConfig: { PIWIK_URL, PIWIK_SITE_ID }
@@ -64,17 +66,21 @@ export default class MyApp extends App {
     const { Component, pageProps } = this.props;
     if (pageProps.statusCode) {
       return (
-        <ThemeProvider theme={theme.colors}>
-          <GlobalStyles />
-          <ErrorPage statusCode={pageProps.statusCode} />
+        <ThemeProvider>
+          <>
+            <GlobalStyles />
+            <ErrorPage statusCode={pageProps.statusCode} />
+          </>
         </ThemeProvider>
       );
     }
     return (
-      <ThemeProvider theme={theme.colors}>
-        <GlobalStyles />
-        <Component {...pageProps} />
-        <CookieConsent />
+      <ThemeProvider>
+        <>
+          <GlobalStyles />
+          <Component {...pageProps} />
+          <CookieConsent />
+        </>
       </ThemeProvider>
     );
   }
--- a/packages/code-du-travail-frontend/src/common/AccessibilityModal.js
+++ b/packages/code-du-travail-frontend/src/common/AccessibilityModal.js
@@ -1,0 +1,51 @@
+import React, { useCallback, useState } from "react";
+import PropTypes from "prop-types";
+import {
+  Heading,
+  InputCheckbox,
+  Modal,
+  Title,
+  theme
+} from "@socialgouv/react-ui";
+
+import { useTheme } from "../layout/ThemeProvider.js";
+
+export const AccessibilityModal = ({ children: renderProp }) => {
+  const [isModalVisible, setModalVisibility] = useState(false);
+
+  const { currentTheme, toggleTheme } = useTheme();
+
+  const openModal = useCallback(e => {
+    e.preventDefault();
+    setModalVisibility(true);
+  }, []);
+
+  const closeModal = useCallback(() => {
+    setModalVisibility(false);
+  }, []);
+
+  return (
+    <>
+      {renderProp(openModal)}
+      <Modal
+        isOpen={isModalVisible}
+        onDismiss={closeModal}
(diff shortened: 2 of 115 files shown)


packages/react-ui/src/field/Checkbox.js
```diff
@@ -68,6 +68,14 @@ const StyledCheckbox = styled.input`
   & {
     visibility: hidden;
   }
+  /* stylelint-disable-next-line */
```
UnbearableBear, line 71: @nkrmr C'est là que j'ai viré le hover

nkrmr: OK top

packages/code-du-travail-frontend/pages/_app.js
```diff
@@ -64,17 +66,21 @@ export default class MyApp extends App {
     const { Component, pageProps } = this.props;
     if (pageProps.statusCode) {
       return (
-        <ThemeProvider theme={theme.colors}>
-          <GlobalStyles />
-          <ErrorPage statusCode={pageProps.statusCode} />
+        <ThemeProvider>
+          <div>
```
lionelB: Fragment are not allowed as theme provider child ?

packages/code-du-travail-frontend/src/common/AccessibilityModal.js
```diff
@@ -0,0 +1,46 @@
...
+} from "@socialgouv/react-ui";
+
+import { useTheme } from "../layout/ThemeProvider.js";
+
+export const AccessibilityModal = props => {
+  const [isModalOpened, setModalOpen] = useState(false);
+  const { currentTheme, toggleTheme } = useTheme();
+  return (
+    <>
+      {React.cloneElement(props.children, {
```
lionelB: It not understood why you need to clone children in the first time. 
A render render prop approach would be simpler, less magic, and less error prone what if I do 
```
<AccessibilityModal onClick={trackAccessibility}>
   <Button>texte dans le bouton</Button>
</AccessibilityModal>
```

packages/code-du-travail-frontend/src/layout/ThemeProvider.js
```diff
@@ -0,0 +1,51 @@
...
+const ThemeContext = React.createContext({
+  currentTheme: colors,
+  toggleTheme: () => {}
+});
+
+export const ThemeProvider = props => {
+  const [currentTheme, setCurrentTheme] = useState(colors);
+
+  useEffect(() => {
+    localStorage.getItem(HIGH_CONTRAST_STORAGE_KEY) === "true" &&
```
lionelB: there is useLocalStorageHook :)

packages/react-ui/src/theme.js
```diff
@@ -28,6 +44,14 @@ export const colors = {
 colors.primaryText = colors.white;
 colors.secondaryText = colors.white;
 
+export const highContrastColors = Object.entries(colors).reduce(
```
lionelB: We could do the same using a level approach were use can choose is level 3 / 7 / 12

lionelB (changes requested):
Cool, just few thinks about localStorage and the Modal approach

packages/code-du-travail-frontend/pages/_app.js
UnbearableBear: The styled component ThemeProvider only accepts a single child and fragment is not accepted

packages/react-ui/src/theme.js
UnbearableBear: In fact I'm going to rename this one BlackAndWhiteTheme and next month we'll add a real highContrast mode (as discussed with clement). We already are in 3 so we only miss a 7 I guess but another time

packages/code-du-travail-frontend/src/layout/ThemeProvider.js
UnbearableBear: I'm going to take a look !

packages/code-du-travail-frontend/pages/_app.js
lionelB: thanks !

packages/react-ui/src/theme.js
lionelB: Oh sorry, the comment was not meant to be actionnable now but more a thought for later

packages/code-du-travail-frontend/src/common/AccessibilityModal.js
UnbearableBear: To be honest I copy pasted the code from `ServiceDeRenseignementsModal` . Also, after taking a look at the modal usages in the project, we do indeed have a `renderProp` usage here:
```
                    <ConventionModal key="convention-modal">
                      {openModal => (
                        <Button variant="secondary" onClick={openModal}>
                          Chercher une convention collective
                        </Button>
                      )}
```
Is that ok if I refactor that the way above ?

lionelB: yup !

packages/code-du-travail-frontend/pages/_app.js
nkrmr: it's works like that for me on another project :

```
    <ThemeProvider theme={theme}>
      <>
        <GlobalStyle />
        <Routes />
      </>
    </ThemeProvider>
```

packages/code-du-travail-frontend/src/layout/ThemeProvider.js
UnbearableBear: Bon, je l'ai utilisé. Je ne suis pas sur d'avoir simplifié le code, au contraire, mais je l'ai utilisé !

lionelB (approved):

packages/code-du-travail-frontend/pages/_app.js
UnbearableBear: Yep because in the end there is no html tag behind Routes and global styles

UnbearableBear: I'm wrong, tested

nkrmr (approved):

PR #2174, opened 2020-01-03
refactor(gitlab): separate package installation

douglasduteil:


lionelB (commented):
I'm not a big fan of splitted gitlab-ci files in each subpackages as it made hard to navigate between files and hard to represent the whole process

.gitlab-ci.yml
```diff
@@ -2,30 +2,97 @@
...
+  stage: "Registration"
+  dependencies: []
+
+#
+
+Create namespace:
+  extends: .base_create_namespace_stage
+  stage: "Registration"
+  except:
+    - /^renovate/*
```
lionelB: this exclude renovate branch / Pr to trigger namespace creation ?

douglasduteil: I will remove this

douglasduteil:
> from @lionelB  
> I'm not a big fan of splitted gitlab-ci files in each subpackages as it made hard to navigate between files and hard to represent the who process

I did it so each package handles it's own process in it's local folder. It's a similar point of view as splitted package.json, Dockerfiles, etc...
If we remove / move a package tomorrow we do not need to think about what to change/remove in over external files that the root `.gitlab-ci.yml` file.

I would like to remove the `.gitlab-ci` folder as end result

douglasduteil:
Outdated

PR #2180, opened 2020-01-06
feat(frontend): home stair tiles

UnbearableBear:
Screenshot 2020-01-06 at 13 05 30
Battle tested on Chrome Firefox Safari and IE11


lionelB (approved):
I wonder if it could be simpler using destructuring and concat
something like 
```js
const [t1, t2, t3,t4, t5,  ...otherThemes] = themes
pyramid = [title, t1, t2, spacerTheme, t3, t4, t5].concat(otherThemes)
```

nkrmr (approved):
💪

UnbearableBear:
Merci Lionel, c'est bien mieux comme ça !

PR #2192, opened 2020-01-07
Fixed input search submit event

nkrmr:
Fixed  IE11 issue to submit event on input search Cf: #1988


UnbearableBear (approved):
IE 11 et ses merdouilles ...

UnbearableBear:
Ca marche bien en local sur ta branche sous ie11, welle done !

lionelB (approved):

lionelB (commented):
I wonder If could simplify it by doing only form.submit() + ie11 event hack

nkrmr:
It's to keep compatibility to old browsers

lionelB:
> It's to keep compatibility to old browsers ? 

we don't support older browser than ie10  I think

PR #2393, opened 2020-02-20
fix(frontend): query is not a regexp

UnbearableBear:
Fix bug when typing a `\` in search
Also, trim query because why not ?


lionelB (approved):
I think it fix some sentry error about renderSuggestion \o/

PR #2441, opened 2020-03-04
feat(data) : add tags in html content WIP

rmelisson:
#1038 
prototype to illustrate enrichment of html content beforehand


rmelisson:
ping @nkrmr

nkrmr (approved):
Top !!  Je m'occupe du reste

PR #2499, opened 2020-03-18
feat(search) add CC ordering based on effectif

rmelisson:
#2223


packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/highlights.test.js.snap
```diff
@@ -3,17 +3,7 @@
 exports[`return highlights matching the slug, correctly hydrated 1`] = `
 Array [
   Object {
-    "breadcrumbs": Array [
```
lionelB: is it ok that breadcrumbs disappear ?

packages/code-du-travail-api/src/server/routes/idcc/idcc.elastic.js
```diff
@@ -1,42 +1,49 @@
 function getIdccBody({ query }) {
+  const matchPhrasePrefix = {
+    match_phrase_prefix: {
```
lionelB: I would prefer leave that inside the query for readbility and I fond it more handy to  copy / paste in kibana too

packages/code-du-travail-api/src/server/routes/idcc/idcc.elastic.js
```diff
@@ -1,42 +1,49 @@
...
     query: {
       bool: {
         filter: [
           {
             term: {
               source: "conventions_collectives"
             }
-          }
+          },
+          matchPhrasePrefix,
```
lionelB: could you explain why you add it to the filter too  ?

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/highlights.test.js.snap
rmelisson: nope, there's probably an issue with the updated data, hence my "not convinced" commit message ;)
need to investigate

packages/code-du-travail-api/src/server/routes/idcc/idcc.elastic.js
rmelisson: Now that we have the `effectif` based rank feature, any convention with effectif would match. 
So we need to ensure we only use effectif on convention matching the query for title or short title. 

But now that I look at it, it means docs _have to_ match the three filters, which is probably wrong.

packages/code-du-travail-api/src/server/routes/idcc/idcc.elastic.js
rmelisson: ok will do, was just trying to avoid duplication

packages/code-du-travail-api/src/server/routes/idcc/idcc.elastic.js
rmelisson: replaced with a should in the filter

lionelB (approved):

PR #2551, opened 2020-03-31
feat(data): add date to Avertissement element

UnbearableBear:
Afin d'être iso avec le site du service public 

Screenshot 2020-03-31 at 15 57 39

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "homepage": "https://github.com/SocialGouv/code-du-travail-numerique#readme",
   "scripts": {
     "build": "lerna run build",
-    "build:ui": "yarn workspace @socialgouv/react-fiche-service-public build && yarn workspace @socialgouv/react-ui build",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": "https://github.com/SocialGouv/code-du-travail-numerique#readme",
   "scripts": {
     "build": "lerna run build",
-    "build:ui": "yarn workspace @socialgouv/react-fiche-service-public build && yarn workspace @socialgouv/react-ui build",
```
lionelB, line 9: wait... its not in alphabetical order :trollface:

UnbearableBear: Ah oui mais là si tu build les fiches sp avant l'ui alors que t'as modifié le deux tu vas avoir des surprises !

lionelB (commented):
parentAttribute feels like a hack to me, so I would add a comment about the context of parentAttribute since its only purpose is to pass dow, a date it should  be refactor later

virginielastisneres (approved):

PR #2573, opened 2020-04-03
ci(gitlab): deploy should not download artifacts by default

douglasduteil:
---

Get the `Register nlp image` artifact only on `Deploy @cdtn/nlp (dev)`


lionelB (changes requested):
It won't work since both api and frontend need to have the correct NLP_URL

lionelB (approved):

PR #2619, opened 2020-04-15
refacotor(data): remove themes index

lionelB:
In order to simplify later refactoring for using datafiller-data, I remove theme index and move it to document index (theme were  already in document index but with less infos)


packages/code-du-travail-frontend/pages/themes/index.js
```diff
@@ -58,14 +59,14 @@ const ThemesPage = ({ pageUrl, ogImage, children = [] }) => (
...
         <Grid>
           {children &&
-            children.map(({ children, icon, slug, title }) => (
+            children.map(({ children, icon, slug, label }) => (
               <Link
                 key={slug}
-                href="/themes/[slug]"
-                as={`/themes/${slug}`}
+                href={`/${getRouteBySource(SOURCES.THEMES)}/[slug]`}
+                as={slug}
```
revolunet, line 65: Isnt the source missing here ?

revolunet (approved):

packages/code-du-travail-frontend/pages/themes/index.js
lionelB: slugs inside breadcrumbs / children arrays now  includes the sources

PR #2638, opened 2020-04-17
Remove accordion tab on directory page

nkrmr:


nkrmr:
Remove on yarn lock issue

PR #2726, opened 2020-05-12
fix(ci): remove frontend waiting nlp initContainer

revolunet:
i dont think this is necessary at frontend level ?


lionelB:
@revolunet i move to api but it's maybe not usefull ? see #2724

revolunet:
good, missed that !
if we dont have any check then the api could be deployed without nlp. may be problematic

PR #2842, opened 2020-06-22
fix: test CI

revolunet:
`before_script` is not executed before `variables` so we have to export `DOCKER_BUILD_ARGS` in the job script directly, just before the docker build

### DIFF
--- a/.gitlab-ci/env.sh
+++ b/.gitlab-ci/env.sh
@@ -52,7 +52,7 @@ fi
 
 #
 
-export API_HOST="api.${DOMAIN}";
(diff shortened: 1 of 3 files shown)


douglasduteil:
Sorry miss click

douglasduteil:
To have the `API_URL` during docker build ?

revolunet:
yes

UnbearableBear:
Merci beaucoup @revolunet !!

lionelB:
another options would be to defined the BUILD_ARGS varaibles inside the before-script rather than variable job's bloc

PR #2953, opened 2020-08-26
fix(rollup): fix rollup config for fiche-sp components

lionelB:
update rollup externals config and replace default export with a named export to fiw rollup warning


revolunet:
C'était cdtn-ui qui était bundled avec?

lionelB:
oui ~~a priori et react via react-uid~~

revolunet (approved):

PR #3001, opened 2020-09-22
bugfix : reco same content #2988

rmelisson:
Fix two things : 
- don't put visited content into its recommendations
- add algo "static" to convention collective recommendations in order to respect generic format

### DIFF
--- a/packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
+++ b/packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
@@ -138,9 +138,7 @@ async function getRelatedItems({ title, settings, slug, covisits }) {
   const filteredItems = covisitedItems
     .concat(searchBasedItems)
     // avoid elements already visible within the item as fragments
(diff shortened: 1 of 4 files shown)


packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
```diff
@@ -138,9 +138,7 @@ async function getRelatedItems({ title, settings, slug, covisits }) {
   const filteredItems = covisitedItems
     .concat(searchBasedItems)
     // avoid elements already visible within the item as fragments
-    .filter((item) =>
```
UnbearableBear, line 141: ohhh la boulette !

UnbearableBear (approved):

packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
rmelisson: ouep, my bad !

PR #3035, opened 2020-10-07
fix(sources): updates typings definitions

lionelB:


packages/sources/index.js
```diff
@@ -70,8 +72,10 @@ const getRouteBySource = (src) => routeBySource[src];
 
 // code-du-travail -> code_du_travail
 /** @type {cdtnSources.getSourceByRouteFn} */
-const getSourceByRoute = (slug) =>
-  sources.find((key) => routeBySource[key] === slug) || slug;
+const getSourceByRoute = (slug) => {
+  const [source] = routes.find(([, value]) => slug === value) || [];
+  return source ? source : null;
```
UnbearableBear, line 77: If think `source | null`is enough here but not sure

lionelB: if you mean `return source || null` ?   it's the same

UnbearableBear (approved):
If it's ok for you it's ok for me

packages/sources/index.js
UnbearableBear: 👌

PR #3065, opened 2020-10-21
feat(related items): remove covisite A/B testing + add log on related card selection 

rmelisson:


rmelisson:
- vérifié sur le Matomo de dev
- en attente de @virginielastisneres sur les modifs UI des "cartes"

packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
```diff
@@ -126,12 +123,11 @@ async function getSearchBasedItems({ title, settings }) {
...
-// get related items, depending on : covisits present & non empty and A/B testing ratio
+// get related items, depending on : covisits present & non empty
 async function getRelatedItems({ title, settings, slug, covisits }) {
-  const useCovisits = Math.random() < abRatio;
-
-  const covisitedItems =
-    covisits && useCovisits ? await getCovisitedItems({ covisits, slug }) : [];
+  const covisitedItems = covisits
+    ? await getCovisitedItems({ covisits, slug })
+    : [];
```
lionelB, line 130: should we add documents (from sem or Es) when covisits don't yet exist ?

rmelisson: yep, we still use search : 
https://github.com/SocialGouv/code-du-travail-numerique/blob/cc2972b987ab73a23fce02cd8b98d425d16f0976/packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js#L135

lionelB (approved):

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -28,6 +28,17 @@ export const RelatedItems = ({ items = [] }) => {
     .filter(({ source }) => !tileSources.includes(source))
     .slice(0, 6);
 
+  const matoSelectRelated = (reco, selection) => {
```
UnbearableBear: slight performance improvement if you put this function outside the component (because the component will get reexecuted at each render)

UnbearableBear (approved):

packages/code-du-travail-api/src/server/routes/items/getRelatedItems.js
```diff
@@ -126,12 +123,11 @@ async function getSearchBasedItems({ title, settings }) {
   );
```
UnbearableBear, line 123: one line above, instead of creating a high order function I would simply use the spread operator:

`.map(({ _source }) => mapSource("search")(_source))` => `.map(({ _source }) => ({..._source, reco: "search"})`

You are less likely to miss a field if we change some (and a -20 lines)

UnbearableBear: (but if the goal is to also filter some fields, nevermind 😄 )

rmelisson: yep, I'm also selecting a subset of the fields

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -1,165 +1,97 @@
...
+    JSON.stringify({
+      reco,
+      selection,
+    }),
+  ]);
+};
 
 export const RelatedItems = ({ items = [] }) => {
-  const tileSources = [SOURCES.EXTERNALS, SOURCES.LETTERS, SOURCES.TOOLS];
+  const nonArticleSources = [SOURCES.EXTERNALS, SOURCES.LETTERS, SOURCES.TOOLS];
```
UnbearableBear: is would use a function called "isArticle" here and then use !isArticle or or isArticle to avoid double negation but it' a minor imrpovement :)

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -1,165 +1,97 @@
...
-                [rootSlug, hash] = slug.split("#");
-              }
-              hash = hash ? `#${hash}` : "";
-              rootSlug = rootSlug ? `/${rootSlug}` : "";
-              const route = getRouteBySource(source);
+                    hash = hash ? `#${hash}` : "";
+                    rootSlug = rootSlug ? `/${rootSlug}` : "";
+                    const route = getRouteBySource(source);
+
+                    href = `/${route}${rootSlug ? `/${slug}` : ""}`;
```
UnbearableBear: no # in href here ?

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -1,165 +1,97 @@
...
-            })}
-          </FlatList>
-        </>
+                  return (
+                    <StyledLinkItem key={slug || url}>
+                      <Link href={href} passHref>
+                        <ArrowLink
+                          rel="nofollow"
+                          arrowPosition="left"
+                          onClick={() => matoSelectRelated(reco, selection)}
```
UnbearableBear: maybe that selection is in fact href and don't need to be in a separate variable

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -1,165 +1,97 @@
...
 
 import { matopush } from "../piwik";
-import { CallToActionTile } from "./tiles/CallToAction";
+
+const matoSelectRelated = (reco, selection) => {
+  matopush([
+    "trackEvent",
+    "selectRelated",
+    JSON.stringify({
+      reco,
```
UnbearableBear, line 20: isn't reco always the same value here since you remove a / b testing ?

packages/code-du-travail-frontend/src/common/RelatedItems.js
UnbearableBear: It might be usefull to keep it !

UnbearableBear (approved):
Minor questions but lgtm !

packages/code-du-travail-frontend/src/common/RelatedItems.js
rmelisson: nope, it can still come from either search or covisits (we use search if no covisits)

packages/code-du-travail-frontend/src/common/RelatedItems.js
rmelisson: So, that's a bit more tricky, in fact `href` was the variable without the `#section` in the next Link `href/as`.
What we miss here is the `as`, I will restore it.

UnbearableBear: Before putting back the "as", are you sure that adding the "#" directly in href will break routing ? I wouldn't be surprised if it worked

packages/code-du-travail-frontend/src/common/RelatedItems.js
```diff
@@ -71,7 +72,7 @@ export const RelatedItems = ({ items = [] }) => {
 
                   return (
                     <StyledLinkItem key={slug || url}>
-                      <Link href={href} passHref>
+                      <Link href={href} as={selection} passHref>
```
UnbearableBear: I think you might be missing the first "/" in the selection variable, but again, it might work without the "as" if you put the "#" i.e. hash directly in the href

PR #3081, opened 2020-10-29
fix(api): use same logic to avoid bugs and reduce complexity

UnbearableBear:
make sure that the logic in getDocumentByUrlQuery is also used in getEsReference later resolution

Screenshot 2020-10-29 at 16 59 13
pas biennn


virginielastisneres (approved):

packages/code-du-travail-api/src/server/routes/search/utils.js
```diff
@@ -46,6 +48,19 @@ const mergePipe = (a, b, max_result) => {
   return mergeDuplicate(res).slice(0, max_result);
 };
 
+const getDataFromUrl = (url) => {
+  const [, sourceRoute, slug] = url.split("/");
+  let source = getSourceByRoute(sourceRoute);
+  if (source === SOURCES.SHEET_MT && !slug.includes("#")) {
```
lionelB, line 57: Does this could lead to problem when there is anchorless sections ?

packages/code-du-travail-api/src/server/routes/search/utils.js
```diff
@@ -46,6 +48,19 @@ const mergePipe = (a, b, max_result) => {
   return mergeDuplicate(res).slice(0, max_result);
 };
 
+const getDataFromUrl = (url) => {
```
lionelB, line 51: what do you think of `extractSourceSlugFromUrl(url)` ?

UnbearableBear: It thought about it but I'm not fond of specifying the returned prop in the function's name. Seemed a bit too specific to me. With `Data` it lets the door open to new url formats without having to change de name (let's say we end up with cdntId in the url); but it's just my opnion.
What about `ExtractDataFromUrl` ?

packages/code-du-travail-api/src/server/routes/search/utils.js
UnbearableBear: Anchorless sections are introductions so there is no problem with that since there will always be at the top of the page

UnbearableBear: (the content in the tiles is the same than the full pages)

lionelB: it fix the problem of invisible display, but to me the problem that we should fix, is why introduction document are not found :/

UnbearableBear: We have the same url for both content so this is inevitable (I think). We have to make a choice when we resolve this. The only other solution I can see is to put the source in the ref but that's what we are already doing in the admin so I guess there's no use in making a refactor right now for something that will "auto-fix" in a near future

UnbearableBear: https://github.com/SocialGouv/code-du-travail-numerique/blob/master/packages/sources/index.js#L34

UnbearableBear: The root issue is that we have the same url for two differents documents. So first, if the fiches were empty, it's not because we don't find them, it's because we had a difference in the ref resolver's logic. We could also decide to always put the introduction instead of the full page when we don't find "#". The issue with that is that we don't always have introductions. Secondly, we decided that we won't put any partial fiches MR in the refs, wether we're in the datafiller or the admin. So they are not supposed to be any SOURCES.SHEET_MT in the future, but they do still are here, because data still has some "#" in refs.

UnbearableBear: the only mecanism which uses the partial fiches MT now is in fact the user's search. Otherwise we deal with entire pages.

lionelB (approved):

lionelB (approved):

PR #3111, opened 2020-11-04
refactor(front): bye bye as

UnbearableBear:
the "as" tag is not necessary anymore !


packages/code-du-travail-frontend/pages/outils/convention-collective.js
```diff
@@ -1,3 +1,28 @@
-import SearchConvention from "../convention-collective";
+import { Container, PageTitle, Section, Wrapper } from "@socialgouv/cdtn-ui";
```
UnbearableBear, line 1: I also moved the CC search page here once and for all.

lionelB:
this will work only in nextjs 10 https://nextjs.org/blog/next-10#automatic-resolving-of-href

UnbearableBear:
Docs say it works since 9.5.3 ? https://nextjs.org/docs/api-reference/next/link

`as: Before Next.js 9.5.3 this was used for dynamic routes, check our previous docs to see how it worked`

packages/code-du-travail-frontend/pages/outils/convention-collective.js
UnbearableBear: (the production bug was due to a wrong "as" prop on Link component)

lionelB (approved):
looks good, need to see why E2E tests failed

UnbearableBear:
> looks good, need to see why E2E tests failed

wrong url for convention collective page 😅

PR #3144, opened 2020-11-18
admin themes and breadcrumbs

UnbearableBear:


packages/code-du-travail-api/src/server/routes/items/index.js
```diff
@@ -54,27 +74,6 @@ router.get("/items/:source/:slug", async (ctx) => {
...
-/**
- * Return document matching the given id.
- *
- * @example
- * http://localhost:1337/api/v1/items/:id
- *
- * @param {string} :id The item id.
- * @returns {Object} Result.
- */
-router.get("/items/:id", async (ctx) => {
```
lionelB: you should refactor to keep this method as it is exactly what you try do do upper.

packages/code-du-travail-data/indexing/cdtnDocuments.js
```diff
@@ -107,38 +147,60 @@ async function* cdtnDocumentsGen() {
...
-    // fix breadcrumbs generation
-    breadcrumbs: getBreadcrumbs(
-      `/${getRouteBySource(SOURCES.SHEET_MT)}/${fiche.slug.replace(/#.*$/, "")}`
-    ),
-    source: SOURCES.SHEET_MT,
-  }));
+  yield splittedFiches.map((fiche) => {
+    const breadcrumbs = getBreadcrumbs(fiche.cdtnId);
+    // we only want cdtnId to be present in full fiches for now
+    delete fiche.cdtnId;
```
lionelB: pourquoi on supprime le cdtnId ?

UnbearableBear: pour pas l'avoir en doublon dans ES, 1 document = 1 cdtnID mais après je veux bien que tu m'explique lundi à quel moment cdtnId se retrouve être l'id du document dans ES parce que là on va avoir un soucis effectivement

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/highlights.test.js.snap
```diff
@@ -19,5 +19,11 @@ Array [
     "source": "fiches_ministere_travail",
     "title": "La démission : comment présenter une démission ?",
   },
+  Object {
```
lionelB: this should not be in the highlight,(it has `isPublished:false`)

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/search.test.js.snap
```diff
@@ -97,17 +97,24 @@ Object {
       "title": "Démission d'un salarié",
       "url": "https://www.service-public.fr/particuliers/vosdroits/F2883",
     },
+    Object {
+      "algo": "pre-qualified",
```
lionelB: same here

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/themes.test.js.snap
```diff
@@ -94,52 +99,52 @@ Object {
...
         },
         Object {
           "label": "Rupture conventionnelle",
-          "slug": "/themes/83-rupture-conventionnelle",
+          "slug": "/themes/rupture-conventionnelle",
         },
         Object {
           "label": "Licenciement : droits des salariés et procédures",
-          "slug": "/themes/84-licenciement-droits-des-salaries-et-procedures",
+          "slug": "/themes/enciement-droits-des-salaries-et-procedures",
```
lionelB: 🤔  is enciement ok ?

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/themes.test.js.snap
```diff
@@ -61,6 +61,11 @@ Object {
       "title": "",
       "url": "/fiche-service-public/indemnite-compensatrice-de-preavis",
     },
+    Object {
+      "source": "external",
```
lionelB: maybe same here (not sure)

packages/code-du-travail-api/src/server/routes/search/getDocumentByUrlQuery.js
```diff
@@ -21,7 +22,11 @@ const getDocumentByUrlQuery = (
     _source,
     query: {
       bool: {
-        filter: [{ term: { slug } }, { term: { source } }],
+        filter: [
+          { term: { slug } },
+          { term: { source } },
+          { term: { isPublished: true } },
```
lionelB: this should not be needed since its used by getEsReference which handle `isPublished`flag

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/search.test.js.snap
```diff
@@ -136,87 +136,17 @@ Object {
...
-      "title": "Peut-on démissionner pendant un congé parental ?",
-      "url": "/fiche-service-public/peut-on-demissionner-pendant-un-conge-parental",
-    },
-    Object {
-      "algo": "pre-qualified",
-      "position": 11,
-      "source": "external",
-      "title": "Démission du salarié à domicile employé par un particulier",
-      "url": "/fiche-service-public/demission-du-salarie-a-domicile-employe-par-un-particulier",
-    },
```
lionelB, line 208: c'est normal que ces fiches aient disparues ?

packages/code-du-travail-api/src/server/routes/search/getDocumentByUrlQuery.js
```diff
@@ -1,3 +1,4 @@
+const { SOURCES } = require("@socialgouv/cdtn-sources");
```
lionelB: import is not used

packages/code-du-travail-api/src/server/routes/search/getEsReferences.js
```diff
@@ -63,6 +63,7 @@ const getEsReferences = async (refs = []) => {
         delete esReference._source.isPublished;
         return esReference;
       }
+      return [];
```
lionelB, line 66: adding it, it will remove document that are not found maybe we should discuss what to do rather  before removing them ?

packages/code-du-travail-api/src/server/routes/themes/search.elastic.js
```diff
@@ -33,7 +29,11 @@ function getThemeQuery({ slug }) {
     ],
     query: {
       bool: {
-        filter: [{ term: { slug } }, { term: { isPublished: true } }],
+        filter: [
+          { term: { slug: slug.replace(/^\d*-/, "") } },
```
lionelB: is this needed since theme slug doesn't contains number now ?

packages/code-du-travail-data/indexing/fetchCdtnAdminDocuments.js
```diff
@@ -112,12 +110,28 @@ export async function getDocumentBySource(source) {
     });
   });
   const docs = await Promise.all(pDocuments);
-  const documents = docs.flatMap((docs) => docs.map(toElastic));
+  const documents = docs.flatMap((docs) =>
+    // /!\ beware /!\
+    // first map should me removed once admin has a way to handle
+    // editorial content's theme / path
+    docs
+      .map((doc) => {
```
lionelB, line 118: why not doing that in toElastic directly ?

packages/code-du-travail-api/src/server/routes/search/getEsReferences.js
UnbearableBear: 100% open to discussion !
Pros: If we find no ref, it "should" mean that that the builded external link will lead to a 404. This is what happens in the api tests on ref on unexisting content. In case this happens, I think the user experience is better if we do not display them.
Cons: We don't see on the frontend the broken ref so we might miss them. But isn't it what we want for the end-user ?

packages/code-du-travail-api/src/server/routes/themes/search.elastic.js
UnbearableBear: Yep to avoid breaking old routes (but I could also do a redirect with next, might be more appropriate)

packages/code-du-travail-api/src/server/routes/__tests__/__snapshots__/search.test.js.snap
UnbearableBear: yes cf discussion plus haut

packages/code-du-travail-data/indexing/fetchCdtnAdminDocuments.js
UnbearableBear: It deserves its own map IMHO

packages/code-du-travail-api/src/server/routes/themes/search.elastic.js
lionelB: old route will be handled on /themes routes. getThemeQuery will search theme documents whose don't have `id` in their slug.

UnbearableBear: I'm not sure I understand. FYI old routes => `themes/45-demission`,  new routes => `themes/demission`

UnbearableBear: how does the old get handled by `themes` routes ? We would get a 404 and that's all I think

packages/code-du-travail-api/src/server/routes/search/getEsReferences.js
lionelB: decision to hide unresolved

lionelB (approved):

packages/code-du-travail-api/src/server/routes/search/utils.js
```diff
@@ -51,14 +51,19 @@ const mergePipe = (a, b, max_result) => {
 const getDataFromUrl = (url) => {
   const [, sourceRoute, slug] = url.split("/");
   let source = getSourceByRoute(sourceRoute);
+  // since theme routing has changed, we will keep this monkey patch for a while
+  let trimmedSlug = slug;
+  if (source === SOURCES.THEMES) {
+    trimmedSlug = slug.replace(/^\d*-/, "");
```
lionelB, line 57: it should be \d+-  not \d* since we only want to match string that match with a number followed by a "-"

UnbearableBear: Yup it's safer you're right. It will never missmatch yet though, because no theme starts with "-" but it could happen one day

packages/code-du-travail-data/indexing/buildThemes.js
```diff
@@ -0,0 +1,56 @@
...
+      id,
+      slug,
+      source,
+      title,
+      document: { icon, description },
+      contentRelations,
+      parentRelations,
+    }) => {
+      const breadcrumbs = getBreadcrumbs(cdtnId);
+      breadcrumbs.pop();
```
lionelB: breadcrumbs: breadcrumbs.slice(0,-1)

PR #3172, opened 2020-11-26
fix(frontend): pretty print documents

UnbearableBear:
✨

### DIFF
--- a/packages/code-du-travail-frontend/src/common/Answer.js
+++ b/packages/code-du-travail-frontend/src/common/Answer.js
@@ -126,4 +126,7 @@ const ShareContainer = styled(Container)`
   @media (max-width: ${breakpoints.mobile}) {
     justify-content: flex-start;
   }
+  @media print {
+    display: none;
+  }
 `;
--- a/packages/code-du-travail-frontend/src/common/Breadcrumbs.js
+++ b/packages/code-du-travail-frontend/src/common/Breadcrumbs.js
@@ -53,6 +53,9 @@ const { breakpoints, fonts, spacings } = theme;
 
 const Nav = styled.nav`
   margin-bottom: ${spacings.small};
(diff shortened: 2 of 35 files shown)


packages/code-du-travail-frontend/src/common/Breadcrumbs.js
```diff
@@ -53,6 +53,9 @@ const { breakpoints, fonts, spacings } = theme;
 
 const Nav = styled.nav`
   margin-bottom: ${spacings.small};
+  @media print {
```
lionelB, line 56: I find it could be useful to have the breadcrumbs printed

packages/code-du-travail-frontend/src/common/Answer.js
```diff
@@ -126,4 +126,7 @@ const ShareContainer = styled(Container)`
   @media (max-width: ${breakpoints.mobile}) {
     justify-content: flex-start;
   }
+  @media print {
```
lionelB, line 129: do you need to remove it since it's also hide in the `<Share Component />` ?

packages/code-du-travail-frontend/src/common/Breadcrumbs.js
UnbearableBear: I think breadcrumbs has a navigation purpose mostly. There already is the main theme displayed, in the printed document, just above the title

packages/code-du-travail-frontend/src/common/Answer.js
UnbearableBear: In the two used, one has a text inside so I'd say yes. Plus it's safer for the print layout too

lionelB (approved):

virginielastisneres (approved):

PR #3190, opened 2020-12-02
fix(ci): allow manual web trigger

revolunet:
According to https://docs.gitlab.com/ee/ci/yaml/#common-if-clauses-for-rules we need `web` source to allow manual trigger from GitLab UI


UnbearableBear (approved):
thanks !

PR #3231, opened 2021-01-05
DO NOT MERGE - DO NOT DELETE 

UnbearableBear:
This branch serves for display / validation purposes


packages/code-du-travail-api/src/server/routes/items/searchBySourceSlug.elastic.js
```diff
@@ -15,7 +15,9 @@ function getSearchBody({ source, slug }) {
       "icon", // outils
       "id",
       "notaHtml",
-      "metaDescription", // dossiers
+      "metaDescription",
+      "populars", // dossiers
```
lionelB, line 19: can you add them to the elastic mappings to ?

packages/code-du-travail-frontend/pages/dossiers/[slug].js
```diff
@@ -46,50 +47,44 @@ function DossierThematique({ dossier }) {
...
-              ))}
-            </Select>
-          </SelectWrapper>
         </Container>
         <MainContainer>
           <FixedWrapper>
             <NavTitle>Sommaire</NavTitle>
             <TableOfContent
-              ids={sortedCategories.map((category) => category.id)}
+              contents={sections.flatMap(({ label, categories }) => [
```
Comment (lionelB) [inline packages/code-du-travail-frontend/pages/dossiers/[slug].js:57]: can you explain the difference with map here, not sure to understand

packages/code-du-travail-frontend/pages/dossiers/[slug].js
```diff
@@ -46,50 +47,44 @@ function DossierThematique({ dossier }) {
...
-                        <DossierLink {...ref} />
-                      </Li>
-                    ))}
-                  </StyledFlatList>
-                </StyledWrapper>
-              ))}
+            {populars.length > 0 && (
+              <PopularWrapper variant="light">
+                <IconStripe centered icon={icons["Populars"]}>
+                  <Title as="h3" stripe="none" id="populaires">
```
Comment (lionelB) [inline packages/code-du-travail-frontend/pages/dossiers/[slug].js]: I think it could be h2, too

packages/code-du-travail-frontend/pages/dossiers/[slug].js
```diff
@@ -106,16 +101,39 @@ DossierThematique.getInitialProps = async ({ query: { slug } }) => {
...
+      )}
+      <StyledFlatList>
+        {visibleRefs.map((ref) => (
+          <Li key={ref.url || ref.externalUrl}>
+            <DossierLink {...ref} />
+          </Li>
+        ))}
+      </StyledFlatList>
+      {!areAllVisible && refs.length > 4 && (
+        <SeeMore small variant="flat" onClick={() => setAllVisible(true)}>
```
Comment (lionelB) [inline packages/code-du-travail-frontend/pages/dossiers/[slug].js]: I guess the`<ViemMore />` component does not fit the case.

lionelB (approved):
Just few comments, but nice work btw !

packages/code-du-travail-frontend/pages/dossiers/[slug].js
Comment (UnbearableBear) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js:57]: `TableOfContent` takes a flat array, without the flat map we would return an arrray of arrays each containing several objects that could have a label key, an id key, or none.

packages/code-du-travail-frontend/pages/dossiers/[slug].js
Comment (UnbearableBear) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js]: Do you prefer that I call it `SeeAll` ? It's only a `Button` so not sure I understand

packages/code-du-travail-frontend/pages/dossiers/[slug].js
Comment (UnbearableBear) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js:57]:
```
type Section = {
  label?: string,
  categories: {
    id: string
  }[],
};

type Contents = {label?: string, id?:string};

const sections : Section[] = [{label:"youhou", categories: [{id: "gdgfzga"}]}];

const contents : Contents[] = sections.flatMap(({ label, categories }) => [
  { label },
  ...categories.map((category) => ({
    id: category.id,
  })),
]);
```
You can check with this in the TS Playground by replacing flatMap with map

Comment (lionelB) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js:57]: thanks it's perfect !

packages/code-du-travail-frontend/pages/dossiers/[slug].js
Comment (lionelB) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js]: Sorry I forget to quote `jsx`
I was refering to the `<ViewMore />` component  which handle "view more" list

Comment (UnbearableBear) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js]: I completly forgot that one ! Maybe that I should add it to the UI lib ?

Comment (lionelB) [reply packages/code-du-travail-frontend/pages/dossiers/[slug].js]: yes why not :)

lionelB:
closed since elasticsearch index goes away

PR #3237, opened 2021-01-08
test: fix e2e test for contrib

lionelB:
update contribution reference value since it has chaged


revolunet (approved):

revolunet:
we may have to add some regexp matching...

lionelB:
> we may have to add some regexp matching...

using a dedicated matcher ? 
not sure to know how to do it but it could be useful for simple regexp validation :)

PR #3270, opened 2021-01-20
feat(search): use internal index for enterprise search WIP

rmelisson:
#3207 
WIP


lionelB:
on peut fermer @rmelisson  ?

PR #3285, opened 2021-02-03
fix: glossary (proposal)

### LINKED ISSUE #3259: Perf: Glossaire
la performance de l'algo de marquage / remplacement des mots du glossaire dans les document prend énormement de temps. Il faudrait voir comment on peut améliorer ses performance / changer l'organisation des pipeline pour éviter que la tache ne dure 1heure

revolunet:
This changes how we process content to highlight glossary terms in the frontend

it uses [hast-util-find-replace](https://github.com/syntax-tree/hast-util-find-and-replace) and [mdast-util-find-replace](https://github.com/syntax-tree/mdast-util-find-and-replace) to find and replace glossary words in the HTML or MDX of our documents. this is done when data is exported to elastic search;

This has some advantages :
 - structure-aware replace : everywhere there's valid text only
 - faster 

And some drawbacks :
 - documents needs valid stucture
 - unist dependency

### About contributions data :

Contributions-data are considered MDX content due to the presence of some custom tags : 

 - `<Tab title="CDI">` use react-ui Accordion
 - `<HDN>` display Hierarchie des normes notice
 - `<Content href="https://www.service-public.fr/particuliers/vosdroits/F317">` : include fiche-vdd react content

At the moment, on the client-side (components/common/Mdx.js), the MDX string content is converted to HTML not MDX because it would need the `unsafe-eval` rule in the CSP to be evaluated ?

demo : https://cf13d01-code-travail.dev2.fabrique.social.gouv.fr/contribution/quelle-est-la-duree-maximale-de-la-periode-dessai-sans-et-avec-renouvellement?q=dur%C3%A9e%20maximale%20de%20la%20p%C3%A9riode%20d%27essai

fix #3259


lionelB:
We need to move this code to cdtn-admin

PR #3288, opened 2021-02-04
ci(gitlab): update to helm@3

douglasduteil:


douglasduteil:
We use kosko now see #3392

PR #3341, opened 2021-03-08
fix(api): specify index while counting prequalified

lionelB:
specify index when counting préqualified


m-maillot (approved):
Good catch !

PR #3346, opened 2021-03-11
feat(infra): elasticsearch module

douglasduteil:


revolunet:
pourquoi un dossier "infra" et pas dans "packages" ?

infra/elasticsearch/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "extends": "@tsconfig/node14/tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "outDir": "lib",
+    "paths": { "@socialgouv/cdtn...infra...elasticsearch/*": ["./src/*"] },
```
m-maillot, line 7: You must set a `baseUrl` if you use `paths` (see https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping)

infra/elasticsearch/package.json
```diff
@@ -0,0 +1,60 @@
...
+          "@babel/preset-typescript"
+        ],
+        "plugins": [
+          "@babel/plugin-transform-modules-commonjs"
+        ]
+      }
+    }
+  },
+  "dependencies": {
+    "@socialgouv/cdtn...dataset...stop-words": "^1.0.0-alpha.1",
```
m-maillot, line 18: I have an issue when I try to run yarn install:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@socialgouv%2fcdtn...dataset...stop-words: Not found".
```
I think it misses documentation, I don't know where I can find this package?

infra/elasticsearch/src/esIndexName.ts
```diff
@@ -0,0 +1,4 @@
+//
+
+export const DOCUMENTS = "documents";
+export const SUGGESTIONS = "suggestions";
```
m-maillot, line 4: We can use namespace here to be more explicit about this variables, like:
```ts
export namespace Indexes {
  const DOCUMENTS = "documents";
  const SUGGESTIONS = "suggestions";
}
```

infra/elasticsearch/src/mapping/suggestion.mapping.ts
```diff
@@ -0,0 +1,23 @@
+export const suggestionMapping = {
```
m-maillot, line 1: Same here, can you explain the goal of this const?

infra/elasticsearch/src/types/Document.ts
```diff
@@ -0,0 +1,129 @@
...
+    readonly url: string;
+  }[];
+
+  readonly shortTitle: string;
+
+  // The local document slug
+  readonly slug: string;
+
+  // Indicates the origin of the document, e.g. 'code_du_travail', 'fiches_service_public' etc.
+  readonly source: string;
```
m-maillot, line 105: We can use an enum here.

infra/elasticsearch/src/types/Document.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  // Indicates wether the document is a split version of anothe one
```
m-maillot, line 107: It's only for contribution and it indicates it's a contribution for a specific CC.

infra/elasticsearch/src/types/Document.ts
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+export interface Document {
```
m-maillot, line 3: I'm not a huge fan of a big flat document. Here, the action is a no-nullish string but for a code-du-travail doc, I have this:
```json
{
		"url": "https://www.legifrance.gouv.fr/affichCodeArticle.do;?idArticle=LEGIARTI000033333034&cidTexte=LEGITEXT000006072050",
		"html": "<p>Les recours hiérarchiques dirigés contre les décisions mentionnées aux articles <a href='/affichCodeArticle.do?cidTexte=LEGITEXT000006072050&idArticle=LEGIARTI000018495600&dateTexte=&categorieLien=cid'>R. 5221-17</a>, <a href='/affichCodeArticle.do?cidTexte=LEGITEXT000006072050&idArticle=LEGIARTI000018495642&dateTexte=&categorieLien=cid'>R. 5221-32 et suivants </a>sont formés auprès du ministre chargé de l'immigration.</p>",
		"dateDebut": 1477958400000,
		"description": "Les recours hiérarchiques dirigés contre les décisions mentionnées aux articles R. 5221-17 , R. 5221-32 et suivants sont formés auprès du ministre chargé de l'immigration",
		"breadcrumbs": [],
		"cdtnId": "232a43a527",
		"excludeFromSearch": false,
		"id": "LEGIARTI000018495604",
		"isPublished": true,
		"metaDescription": "Les recours hiérarchiques dirigés contre les décisions mentionnées aux articles R. 5221-17 , R. 5221-32 et suivants sont formés auprès du ministre chargé de l'immigration",
		"slug": "r5221-19",
		"source": "code_du_travail",
		"text": "Les recours hiérarchiques dirigés contre les décisions mentionnées aux articles R. 5221-17 , R. 5221-32 et suivants sont formés auprès du ministre chargé de l'immigration.\n",
		"title": "R5221-19"
	}
```
There is no action. So action can be Nullable.

I think we can split by document source, something like that:
```ts 
interface FicheServicePublic {
   // FicheServicePublic specific fields
    data: string;
}

interface Contribution {
   // Contribution specific fields
    legalReferences: string;
}

enum Source {
    CONTRIBUTION = "contribution",
    FICHE_SP = "fiche_service_public",
} 

type Doc = {
   // Common fields 
    cdtn_id: string;
    source: Source;
} & (FicheServicePublic | Contribution)

const data: Doc[] = [{
    "cdtn_id": "ID1",
    "source": "fiche_service_public",
    "data": "La donnée"
}, {
    "cdtn_id": "ID2",
    "source": "contribution",
    "legalReferences": "Références légales"
}];

data.forEach((item) => {
    console.log("ID: ", item.cdtn_id);
    switch (true) {
        case item.source === Source.FICHE_SP:
        console.log("Item: ", (item as FicheServicePublic).data);
        break;
        case item.source === Source.CONTRIBUTION:
        console.log("Item: ", (item as Contribution).legalReferences);
        break;
    }
})
```

infra/elasticsearch/src/types/index.ts
```diff
@@ -0,0 +1,3 @@
+//
+
+export * from "./Document";
```
m-maillot, line 3: Why does this file start with an uppercase?

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+  indexName,
+  mappings,
+}: UtilsOptions & { mappings: Mappings }): Promise<void> {
+  const { body } = await client.indices.exists({ index: indexName });
+  if (body) {
+    try {
+      await client.indices.delete({ index: indexName });
+      logger.info(`Index ${indexName} deleted.`);
+    } catch (error: unknown) {
+      logger.error("index delete", error);
```
m-maillot, line 36: I think we can raise the error here. If we can't delete the existing indices, why should we continue?

m-maillot: Perhaps improve the log with an action: `Failed to  delete index...`

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+          },
+          number_of_replicas: 1,
+          number_of_shards: 1,
+        },
+      },
+      index: indexName,
+    });
+    logger.info(`Index ${indexName} created.`);
+  } catch (error: unknown) {
+    logger.error("index create", error);
```
m-maillot, line 65: Perhaps improve the log with an action: `Failed to create index...`

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    throw error;
+  }
+}
+
+export async function version({ client }: UtilsOptions): Promise<void> {
+  const { body } = await client.info<{ version: { number: string } }>();
+  logger.info(body.version.number);
```
m-maillot, line 72: I'm not expected to log the version here. We should either rename the function to `logVersion` or return the version and let the caller do what he wants with the version.

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+      items: { index: { status: number } }[];
+    }
+
+    const body = documents.flatMap((doc, i) => [
+      {
+        index: {
+          _index: indexName,
+          // if available, use our cdtnId as the actual Elastic document id
+          ...(doc.cdtnId && { _id: doc.cdtnId }),
+          // unless we're in testing mode where we use position
```
m-maillot, line 92: I think it's not a good idea to have a specific code for testing.  Moreover, I didn't see any test on this method that required this code.

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const resp = await client.bulk<BulkApiResponse>({
+      body,
+      index: indexName,
+    });
+    if (resp.body.errors) {
+      const errorDocs = resp.body.items.filter(
+        (item) => item.index.status != 201
+      );
+      logger.error(`Errors during indexation : ${JSON.stringify(errorDocs)}`);
```
m-maillot, line 107: I think we can throw an error here.

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    logger.info(`Index ${documents.length} documents.`);
```
m-maillot, line 109: Can be useful to return this information (count of documents indexed) in place of a void.

infra/elasticsearch/src/utils/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    format: "json",
+  });
+
+  const IndicesToDelete = getIndicesToDelete(patterns, timestamp, indices);
+  const pIndicesToDelete = IndicesToDelete.map(async ({ index }) =>
+    client.indices.delete({ index })
+  );
+
+  return Promise.all(pIndicesToDelete).then(() => {
+    logger.info(`Remove ${pIndicesToDelete.length} old indices`);
```
m-maillot, line 143: It can be useful to return all indices deleted.

infra/elasticsearch/src/mapping/.eslintrc.yml
```diff
@@ -0,0 +1,6 @@
+rules:
+  # NOTE(douglasduteil): disable naming-convention for elasticsearch mapping
+  # As the mapping files are following the elasticsearch explicit mapping format
+  # https://www.elastic.co/guide/en/elasticsearch/reference/current/explicit-mapping.html
+  # we do not want to enfore a external naming convention here.
+  "@typescript-eslint/naming-convention": "off"
```
m-maillot, line 6: If you disable the rule here, you don't need the comment `/* eslint-disable @typescript-eslint/naming-convention */` in the code. I miss something?

m-maillot (commented):
I'm a huge fan of typescript, I'm 100% for this PR 👍 
I suggest some changes about types, especially for the documents in ES.

I saw after a few minutes, the code was copy/paste from existing code, so you have some comment about existing code, sorry 😄 

I think the process to populate the ES should be in the admin project. So, this code should be in the admin project. The type of documents in the ES, should be in a specific package that can be used by the frontend and the admin. The code to create/index/... should not be reachable by the frontend. What do you think?

infra/elasticsearch/src/mapping/suggestion.mapping.ts
lionelB: it's the es mapping for the suggestion index

infra/elasticsearch/package.json
lionelB: we could inline the stopword in this module

lionelB:
> I think the process to populate the ES should be in the admin project. So, this code should be in the admin project. The type of documents in the ES, should be in a specific package that can be used by the frontend and the admin. The code to create/index/... should not be reachable by the frontend. What do you think?

to avoid massie refactoring, I hase suggested to extract the code that update es index into a module (so only  move / rename files) and in a second step, add the code into a cdtn-admin modules, publish it to npm and in a theird step, replace the local module by the pulblished one.

>  The code to create/index/... should not be reachable by the frontend.

not sure to understand, why. The cdtn will not have the credentials to update prod, but will use the code to create local index to run e2e test for the api

douglasduteil:
Perfer #3359  for now

PR #3359, opened 2021-03-17
refactor(infra): elasticsearch module

douglasduteil:
---

Partial extract from

```
git checkout douglasduteil/refactor-data-extract-the-elasticsearch-infra-domain -- infra/elasticsearch/
```

Linked to #3336


packages/code-du-travail-data/Dockerfile
```diff
@@ -17,8 +17,6 @@ WORKDIR /app
 
 COPY ./package.json /app/package.json
 COPY --from=cdtn-base-image /app/packages/code-du-travail-data/dist /app/dist
-COPY ./dataset/stop_words/stop_words.json ./dataset/stop_words/stop_words.json
-COPY ./dataset/synonyms/synonyms.json ./dataset/synonyms/synonyms.json
```
lionelB, line 21: should you `COPY infra/elasticsearch/package.json` ?

lionelB (commented):
Thanks for the PR . having few questions
- why convert code to commonjs 
- do we need to .map files ?
- it would be better to have code placed in `src` rather that `lib` to follow the project convention
- vectorizer seems to be transpiled, is it ok ?
- I would place the DOCUMENTS and SUGGESTION in an objet esIndexName 
```diff
- import {documentsMapping, suggestionMapping, createIndex, DOCUMENTS SUGGESTIONS} from "@socialgouv/cdtn-elasticsearch"
+ import {documentsMapping, suggestionMapping, createIndex, esIndexName} from "@socialgouv/cdtn-elasticsearch"
```
- I would also export some helper methods like createDocumentIndex / createSuggestionIndex so you don't need to mappings / indexBaseName

packages/code-du-travail-data/Dockerfile
douglasduteil: done above ligne 19 :)

douglasduteil:
>     * why convert code to commonjs

The best compatibility on node execution context is still commonjs

>     * do we need to .map files ?

No we don't they will be removed in the tsify PR

>     * it would be better to have code placed in `src` rather that `lib` to follow the project convention

I place the code in the lib folder so no changes are made in the entrypoint between this PR and the tsify one

>     * vectorizer seems to be transpiled, is it ok ?

I can remove some of the transpilation noise or restore the source file

>     * I would place the DOCUMENTS and SUGGESTION in an objet esIndexName

The external API can change in the future PRs 

>     * I would also export some helper methods like createDocumentIndex / createSuggestionIndex

The external API can change in the future PRs

packages/code-du-travail-data/Dockerfile
lionelB: sorry i don't understand but l19 you only copy dist from data not the infra ?

douglasduteil: Ah sorry I looked at wrong file 😓

douglasduteil: This Dockerfile uses a bundled version of the sources. All dependencies are in the `dist` folder. We don't have to manually deal with them.

douglasduteil: > should you `COPY infra/elasticsearch/package.json` ?

The `COPY infra/elasticsearch/package.json` is in the "main" Dockerfile which builds the sources

lionelB (approved):

PR #3362, opened 2021-03-17
refactor(data): use ingest in docker data image

douglasduteil:


lionelB:
will be done after ts refactoring

PR #3370, opened 2021-03-17
refactor(infra): tsify infra/elasticsearch with standalone pipeline

douglasduteil:


infra/elasticsearch/src/vectorizer/index.ts
```diff
@@ -1,22 +1,34 @@
...
+  if (response.ok) {
+    const json: { outputs: number[][] } = await response.json();
+    return json.outputs;
+  } else {
+    throw new Error(response.statusText);
+  }
+}
+
+export function preprocess(text: string): string {
+  const stopWords = new Set(stopwords.map(stripAccents));
```
lionelB: this should be outside of the function there is no need to recreate the stopwords list

lerna.json
```diff
@@ -7,9 +7,6 @@
     }
   },
   "npmClient": "yarn",
-  "packages": [
-    "packages/*"
```
lionelB, line 11: is it ok to remove les package path for lerna ?

infra/elasticsearch/src/index.ts
```diff
@@ -0,0 +1,6 @@
+//
```
lionelB: not needed

infra/elasticsearch/src/client.ts
```diff
@@ -0,0 +1,188 @@
...
+    throw error;
+  }
+}
+
+export async function indexDocumentsBatched({
+  client,
+  indexName,
+  documents,
+  size = 1000,
+}: UtilsOptions & { documents: Document[]; size: number }): Promise<void> {
```
lionelB, line 121: should it be it simpler to split the arguments ?

infra/elasticsearch/src/client.ts
```diff
@@ -0,0 +1,188 @@
...
+  const pIndicesToDelete = IndicesToDelete.map(async ({ index }) =>
+    client.indices.delete({ index })
+  );
+
+  return Promise.all(pIndicesToDelete).then(() => {
+    logger.info(`Remove ${pIndicesToDelete.length} old indices`);
+  });
+}
+
+export function* chunks<T = unknown>(
```
lionelB, line 147: is ti a best practice to give a Generic an `unknonw` type ?

infra/elasticsearch/src/vectorizer/index.ts
douglasduteil: I would argue that less globales is better but ok.
I will restore the line 👍

lerna.json
douglasduteil: lerna is directly using the workspace described in the `package.json`
This change could be a separate pr

infra/elasticsearch/src/index.ts
douglasduteil: my aesthetic habits ... 😏

infra/elasticsearch/src/client.ts
douglasduteil: What do you mean ?

infra/elasticsearch/src/client.ts
douglasduteil: it's more like a temporally workaround seems I do not know what is the type to put heer

m-maillot:
Can we close it? `infra/elasticsearch` has been moved to `cdtn-admin` project which is currently move to TS.
@douglasduteil @lionelB

douglasduteil:
Yeap @m-maillot

PR #3385, opened 2021-03-20
style(frontend): lint the tests and mocks folder too

douglasduteil:
---

```sh
$ yarn --cwd packages/code-du-travail-frontend lint-js --fix
yarn run v1.22.10
$ eslint src pages __tests__ __mocks__ --fix
Done in 11.29s.
```


douglasduteil:
Note to self:
- The frontend server is not linted and some console.log make eslint not happy
- We should align the `.eslintignore` to eslint can be run with `eslint .`

lionelB:
for the console log we could use the cdtn-logger

lionelB (approved):

PR #3416, opened 2021-04-01
feat(simulator): init folder for publicode models

m-maillot:
I created a specific package for the publicodes models. 
The idea is to separate models from the UI code. Later, we can move it in a specific repo and populate the frontend from the admin (like fiche SP, kalidata...), for instance.

I create a project with typescript. The idea is to move our javascript to typescript code (like admin).

This package will contain models but also unit tests to validate models. We can add rules to our models quickly with no fear (or a little bit).

I don't have:
 * implement the solution to link this package to the frontend (the idea is to use the frontend API to provide the models)
 * implement linter, CI,... I think it will be more efficient if a more advanced developer can help me (pair programming session ? :) )
 * implement the whole model for computing retirement notice.


packages/code-du-travail-modeles/src/utils/merger.ts
```diff
@@ -0,0 +1,25 @@
...
+      const fullpath = path.join(dirPath, filename);
+      if (fs.statSync(fullpath).isDirectory()) {
+        return concatenateFilesInDir(fullpath);
+      } else {
+        return filename.endsWith(".yaml")
+          ? fs.readFileSync(fullpath).toString()
+          : "";
+      }
+    })
+    .reduce((acc, cur) => acc + "\n" + cur, "");
```
lionelB:
```suggestion
    .join("\n");
```

lionelB (approved):

PR #3425, opened 2021-04-08
feat(simulator): add origin & seniority steps for business retirement simu

m-maillot:
First implementation of the steps: origine, ancienneté and résultat.

It's a first run. The idea is to implement the publicodes then come back on screens to improve them (with the feedbacks of the team).

 * Origine: I follow the github issue but I think the text can be better. The idea: submit to the team to get feedbacks.
 * Ancienneté: The idea is to ask the ancienneté in month or with 2 dates. In this first version, I'll ask only for the ancienneté in month. 
 * Résultat: Keep it simple first. Some information will be show to user depending of the computation but it will be provide by the publicodes.

CleanShot 2021-04-08 at 14 53 58@2x

CleanShot 2021-04-08 at 14 54 13@2x

CleanShot 2021-04-08 at 14 54 37@2x

So next step: Implement the publicodes context 🙃

### DIFF
--- a/packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Anciennete.tsx
+++ b/packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Anciennete.tsx
@@ -1,0 +1,20 @@
+import React from "react";
+
+import { TextQuestion } from "../../common/TextQuestion";
+import { isPositiveNumber } from "../../common/validators";
+
+function AncienneteStep() {
+  return (
+    <>
+      <TextQuestion
+        name="contrat salarié - ancienneté"
+        label="Quelle est votre ancienneté dans l’entreprise en mois&nbsp;?"
+        inputType="number"
+        validate={isPositiveNumber}
+        placeholder="0"
+      />
+    </>
+  );
+}
+
+export { AncienneteStep };
--- a/packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Origine.tsx
+++ b/packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Origine.tsx
@@ -1,0 +1,49 @@
+import { InputRadio } from "@socialgouv/cdtn-ui";
+import React from "react";
+import { Field } from "react-final-form";
+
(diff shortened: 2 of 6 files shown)


packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Anciennete.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Field } from "react-final-form";
+
+import { InlineError } from "../../common/ErrorField";
+import { Question } from "../../common/Question";
+import { isPositiveNumber } from "../../common/validators";
+
+function AncienneteStep() {
+  return (
+    <>
+      <Question as="p" required>
```
lionelB: this would be better to use a `label` tag rather than a `p` so user can click on label to focus the input

lionelB: you could use A TextQuestion Component and precise it's a `type="number"`

packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Origine.tsx
```diff
@@ -0,0 +1,53 @@
...
+              label="L’employeur décide de lui-même de mettre à la retraite le salarié par une décision adressée à celui-ci."
+              {...props.input}
+            />
+          )}
+        </Field>
+        <Field
+          type="radio"
+          name="contrat salarié - mise à la retraite"
+          value="non"
+          id="depart-retraite"
```
lionelB: not sure id is a prop of `<Field>` component

m-maillot: I deleted them. I don't know why I put an `id` here 🤷

packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Anciennete.tsx
m-maillot: Good idea, I change it 👍

lionelB (approved):

PR #3426, opened 2021-04-08
feat(frontend): add a11y navigation skip link

lionelB:
use tab to show the links

### DIFF
--- a/packages/code-du-travail-frontend/pages/index.js
+++ b/packages/code-du-travail-frontend/pages/index.js
@@ -54,12 +54,11 @@ const Home = ({ themes = [], highlights = [] }) => (
       description="Posez votre question sur le droit du travail et obtenez une réponse personnalisée à vos questions (contrat de travail, congés payés, formation, démission, indemnités)."
     />
     <SearchHero />
-    <div id="content">
-      {highlights.length > 0 && (
-        <Highlights highlights={highlights.slice(0, 4)} />
-      )}
-      {themes.length > 0 && <Themes themes={themes} />}
-    </div>
+    {highlights.length > 0 && (
+      <Highlights highlights={highlights.slice(0, 4)} />
+    )}
+    {themes.length > 0 && <Themes themes={themes} />}
+
     <Section>
       <Container>
         <PageTitle
--- a/packages/code-du-travail-frontend/pages/index.js
+++ b/packages/code-du-travail-frontend/pages/index.js
@@ -54,10 +54,12 @@ const Home = ({ themes = [], highlights = [] }) => (
       description="Posez votre question sur le droit du travail et obtenez une réponse personnalisée à vos questions (contrat de travail, congés payés, formation, démission, indemnités)."
     />
     <SearchHero />
-    {highlights.length > 0 && (
-      <Highlights highlights={highlights.slice(0, 4)} />
-    )}
-    {themes.length > 0 && <Themes themes={themes} />}
+    <div id="content">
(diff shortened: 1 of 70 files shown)


packages/code-du-travail-frontend/pages/index.js
```diff
@@ -54,10 +54,12 @@ const Home = ({ themes = [], highlights = [] }) => (
       description="Posez votre question sur le droit du travail et obtenez une réponse personnalisée à vos questions (contrat de travail, congés payés, formation, démission, indemnités)."
     />
     <SearchHero />
-    {highlights.length > 0 && (
-      <Highlights highlights={highlights.slice(0, 4)} />
-    )}
-    {themes.length > 0 && <Themes themes={themes} />}
+    <div id="content">
```
m-maillot: From the rule 12.6:

> La zone possède un rôle WAI-ARIA de type landmark correspondant à sa nature ;

From the landmark page:

> La zone de contenu principal doit avoir un attribut WAI-ARIA role="main" ;

I think we should set `main` in place of `content` here.

packages/code-du-travail-frontend/pages/index.js
```diff
@@ -54,10 +54,12 @@ const Home = ({ themes = [], highlights = [] }) => (
...
+      {highlights.length > 0 && (
+        <Highlights highlights={highlights.slice(0, 4)} />
+      )}
+      {themes.length > 0 && <Themes themes={themes} />}
+    </div>
```
m-maillot: Why the tools are not in the section?

packages/code-du-travail-frontend/src/layout/Header/SkipNavLink.js
```diff
@@ -0,0 +1,39 @@
...
+  top: 0;
+  left: 0;
+`;
+
+const Li = styled.li`
+  padding-left: 0;
+`;
+const { spacings } = theme;
+const SkipLink = styled.a`
+  background-color: ${({ theme }) => theme.paragraph};
```
m-maillot, line 36: I think we should use another color :/

CleanShot 2021-04-09 at 10 38 53@2x

The contrast is poor with this link color.

packages/code-du-travail-frontend/src/layout/Layout.js
```diff
@@ -5,14 +5,21 @@ import styled, { css } from "styled-components";
...
     <BackgroundContainer>
+      <SkipNavLink />
       <BackgroundLayer currentPage={currentPage} />
       <Header currentPage={currentPage} />
       <ErrorBoundary message="Une erreur est survenue">
-        <StyledMain>{children}</StyledMain>
+        {isHome ? (
+          <StyledMain>{children}</StyledMain>
+        ) : (
+          <StyledMain id="content">{children}</StyledMain>
```
m-maillot: Same here, it should be `main` in place of `content`.

m-maillot (changes requested):
The navigation skip link doesn't work on the whole site.
For instance, if I open this page: https://9805866-code-travail.dev2.fabrique.social.gouv.fr/outils/preavis-demission
I don't have access to the navigation skip.

It's the same if I navigate in the site. But I don't know if it's required in the case 🤷

packages/code-du-travail-frontend/src/layout/Header/SkipNavLink.js
lionelB: replace with white

m-maillot (approved):

PR #3436, opened 2021-04-13
fix(k8s): use code.travail.gouv.fr as default

revolunet:
this makes code.travail.gouv.fr the default ingress rule.

The ingress-prod.yml ensure a 301 redirect to code.travail.gouv.fr when another host is used


lionelB (approved):

douglasduteil (approved):
K

PR #3438, opened 2021-04-14
fix(frontend): replace getStaticPath with getServerSideProps

lionelB:
replace getStaticPath with getServerSideProps since we can't rely on build for prod and dev


douglasduteil (approved):
Yes

PR #3442, opened 2021-04-16
ci: fix production pipeline job

lionelB:
should fix ref problem when when click  "production" on tag pipeline

### DIFF
--- a/.gitlab-ci/stages/manual.yml
+++ b/.gitlab-ci/stages/manual.yml
@@ -49,9 +49,9 @@ Production:
       # that only bump version / udpate changelog
       - $CI_COMMIT_MESSAGE =~ /^chore(release):*/
   script:
-    - echo "deploy ${COMMIT_TAG}"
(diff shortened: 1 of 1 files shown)


.gitlab-ci/stages/manual.yml
```diff
@@ -49,9 +49,9 @@ Production:
       # that only bump version / udpate changelog
       - $CI_COMMIT_MESSAGE =~ /^chore(release):*/
   script:
-    - echo "deploy ${CI_COMMIT_REF_SLUG}"
+    - echo "deploy ${COMMIT_TAG}"
```
m-maillot: I think it's `CI_COMMIT_TAG`, cf: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

lionelB: good catch ! (COMMIT_TAG exist only after env.sh have run)

m-maillot (approved):

PR #3446, opened 2021-04-19
feat(simulator): add the labor agreement for plasturgie

m-maillot:
Add the labor agreement rules of plasturgie to publicodes models.


douglasduteil (approved):
Superbe

PR #3451, opened 2021-04-23
fix(accessibility): fix focus on CC tooltip

m-maillot:
Fix the focus with the CC tooltip

CleanShot 2021-04-23 at 15 28 45

Related to #3405


douglasduteil (approved):
🌈 Coool 🌈

PR #3454, opened 2021-04-27
feat(react-ui): add new svg icons

lionelB:
add new themtic file icons


m-maillot (approved):
Look good for me.
What is the goal of these new SCG icons?

lionelB:
these icons will be used in the new thematic file

PR #3494, opened 2021-05-19
feat(simulator): show references on retirement notice simulator

m-maillot:
Show the references on the result page. It's the same idea as the notification. I put some tests because it's important information. 

I fixed also the typo on `profressional`.

Related to #3268


packages/code-du-travail-frontend/src/outils/DureePreavisRetraite/steps/Result.tsx
```diff
@@ -17,6 +18,9 @@ function ResultStep({ form }: WizardStepProps): JSX.Element {
   }, [form]);
 
   const notifications = publicodesContext.getNotifications();
+  const references = publicodesContext.getReferences();
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
m-maillot: Not needed anymore, should be removed. It was here for the `A11yLink` component which has been move to typescript.

packages/code-du-travail-frontend/src/common/A11yLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+import React from "react";
+
+type Props = {
+  children: string;
+} & React.DetailedHTMLProps<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  HTMLAnchorElement
+>;
+
+export const A11yLink = React.forwardRef<HTMLAnchorElement, Props>(
```
m-maillot, line 11: I'm not sure the `forwardRef` is used in the code. I didn't move it to a classic React component but I think we can do. What do you think ?

lionelB: it's needed for internal link that are wrapped inside a `<Link>` tag

packages/code-du-travail-modeles/src/__test__/depart.mise.retraite.industrie.pharmaceutique.references.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        "contrat salarié . ancienneté": 6,
+        "contrat salarié . mise à la retraite":
+          retirement === "mise" ? "oui" : "non",
+        "contrat salarié . convention collective . industrie pharmaceutique . conclu après 1 juillet 2019":
+          "oui",
+        "contrat salarié . convention collective . industrie pharmaceutique . groupe": 5,
+      })
+    );
+
+    expect(result).toEqual(expect.arrayContaining(expectedReferences));
```
lionelB, line 55: maybe we could use inline snapshots to avoid manually maintaining references

packages/code-du-travail-modeles/src/utils/GetReferences.ts
```diff
@@ -0,0 +1,25 @@
...
+export type References = {
+  article: string;
+  url: string;
+};
+
+export function getReferences(engine: Engine): Array<References> {
+  return Object.values(engine.getParsedRules())
+    .filter(
+      (rule) =>
+        rule.rawNode.références && !!engine.evaluate(rule.dottedName).nodeValue
```
lionelB: not very comfortable with the double `!!`  
what are the possible value of nodeValue when it is truthy  or falsy

packages/code-du-travail-frontend/src/common/A11yLink.tsx
m-maillot: Are you sure? I didn't find any references in the code when searching `A11yLink`.

packages/code-du-travail-modeles/src/__test__/depart.mise.retraite.industrie.pharmaceutique.references.spec.ts
m-maillot: It doesn't work with TDD :/ Snapshot can be useful but I think it's important data and it should be checked with precaution.

packages/code-du-travail-modeles/src/utils/GetReferences.ts
m-maillot: The nodeValue contains: `number | boolean | string | Record<string, unknown>`

Here we can use `nodeValue !== false` in place of `!!`.

m-maillot: I changed to 
```js
rule.rawNode.références && engine.evaluate(rule.dottedName).nodeValue !== null
```
and all tests passed :+1:

lionelB: thx

lionelB (approved):

PR #3747, opened 2021-09-15
fix: try to simplify docker stuff

revolunet:
Attempt to remove the docker base image which is used by frontend and api in order to have separate, independently buildable containers.

One issue is that the "base" image is used both for CI tooling and frontend container

Todo :
 - reduce frontend container size
 - reduce frontend container image complexity


maxgfr:
@revolunet Hello, je pense qu'on peut close la PR ?

PR #3840, opened 2021-10-11
fix(events): debounce cc search events - 3814

rmelisson:
#3814

### DIFF
--- a/packages/code-du-travail-frontend/src/outils/ConventionCollective/index.tsx
+++ b/packages/code-du-travail-frontend/src/outils/ConventionCollective/index.tsx
@@ -97,22 +97,29 @@ function AgreementSearchTool({ icon, title }: Props): JSX.Element {
   );
 }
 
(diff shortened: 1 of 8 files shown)


maxgfr (commented):
👍

packages/code-du-travail-frontend/src/outils/ConventionCollective/index.tsx
```diff
@@ -97,22 +97,29 @@ function AgreementSearchTool({ icon, title }: Props): JSX.Element {
   );
 }
 
-const AgreementSearchWithContext = (props: Props): JSX.Element => {
+const AgreementSearchUI = (props: Props) => {
```
m-maillot:
```suggestion
const AgreementSearchUI = (props: Props): JSX.Element => {
```

m-maillot (approved):
:+1:

PR #4599, opened 2022-10-13
fix(ci): deprecated `set-output` fixed

maxgfr:
cf : [github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)


m-maillot (approved):
Bien vu !

PR #7200, opened 2026-03-27
fix(events): suppression de l'event qui track l'input `agreement search`

### LINKED ISSUE #7196: Supprimer l'event cc_search
Supprimer les events cc_search du site car il envoie trop de données

maxgfr:
fix #7196


m-maillot (approved):
Nickel 👌

